PR #7, opened 2016-03-24
Add tests for ds9 region parser

joleroi:
This is a continuation of https://github.com/astropy/regions/pull/5
Addressing the 'correctness test'

* [x] Add functionality to write regions to ds9
* [x] Manually compare to input files an check everything is ok
* [x] Copy ds9 output to reference file and use for future testing


joleroi:
@keflavich I added tests that read the fk5, galactic, and physical input files and creates reference files (checked by hand). I guess that is sufficient to start playing with the parsers. What do you think?

keflavich:
Yes, the testing structure looks good.  Any other interesting failure points along the way?

joleroi:
@keflavich 
The test failure seems related to older astropy versions (that didn't have ecliptic coordinates?), I can try to have a look but I guess it's more efficient if someone who knows astropy does that.

keflavich:
@joleroi Looks like you're right.  We should ask @mwcraig about the astropy versions linked to older numpy versions on conda: they seem on the old side (1.0.3<->1.8, 1.0.4<->1.9).

However, the solution is to do a version check in the testing and use
`@pytest.mark.xfail(astropy.version < 1.(something))`
where the `version` strings should be wrapped in `StrictVersion` from `distutils`.

joleroi:
One test fixed -> 2 others fail :grin:

Seem like StrictVersion does not like the astropy dev version strings ...

keflavich:
Ah, that's true... maybe we need `LooseVersion` or `.strip(".dev")` or something.

joleroi:
Checks pass

cdeil:
Would it be possible to merge this today so that people (specifically me) can start playing with this?
Further work can always be done in future PRs, no?

keflavich:
Yes, reviewing.

astrofrog:
:+1:

PR #8, opened 2016-03-25
Plotting circles

joleroi:
Example for plotting astropy.regions

@keflavich, @astrofrog 
* I don't know what API you had in mind (to_mpl()/plot() on each object vs. object_to_mpl(region)
* I don't know how you want to do the plotting (transformation to pixels in astropy.regions then convert to patch vs. let matplotlib do the conversion (like I did for now) downside - always have to create wcsaxis instance)

To run an example go to  ``regions/plotting`` and run 
```
python plotting.py
```


regions/plotting/plotting.py
```diff
@@ -0,0 +1,58 @@
...
+           "<class 'regions.shapes.circle.CirclePixelRegion'>": 'pixcircle',
+           "<class 'regions.shapes.ellipse.EllipseSkyRegion'>": 'skyellipse',
+           "<class 'regions.shapes.ellipse.EllipsePixelRegion'>": 'pixellipse',
+           "<class 'regions.shapes.polygon.PolygonSkyRegion'>": 'skypolygon',
+           "<class 'regions.shapes.polygon.PolygonPixelRegion'>": 'pixpolygon'}
+
+    output = list()
+
+    for reg in region_list:
+        temp = str(reg.__class__)
```
keflavich: you could just use the region class directly, rather than the string name of the class

regions/plotting/plotting.py
```diff
@@ -0,0 +1,58 @@
...
+    patch = mpatches.Circle(center, **kwargs)
+
+    return patch
+
+if __name__ == "__main__":
+    from astropy.io import fits
+    from astropy.wcs import WCS
+    import matplotlib.pyplot as plt
+
+    hdulist = fits.open('tests/empty_map.fits')
```
keflavich: instead of adding the empty_map to the repository, why not add the header and generate the empty_map data on the fly?

keflavich:
For API, I like `to_mpl_patch` for single regions, and `to_mpl_patchcollection` for multiple regions.  It would be nice if we could have some sort of overloading that would allow `matplotlib.pyplot.plot(region)`, but that is probably unrealistic.  However, supporting `matplotlib.axes.Axes.add_artist` should be possible.

astrofrog:
+1 to ``to_mpl_patch`` and ``to_mpl_patchcollection``

joleroi:
What about plotting SkyRegions transformation? Ok to use WCSaxis?

keflavich:
yes, that's the right thing to do in examples, but that doesn't need to be part of the API (as long as the API is readily usable with wcsaxes)

joleroi:
I don't understand. Let met explain my problem again. This is how it's done in gammapy

```
def to_mpl_artist(self, ax, **kwargs):
        """Convert to mpl patch using a given wcs transformation
        Parameters
        ----------
        ax : `~astropy.wcsaxes.WCSAxes`
            WCS axis object
        kwargs : dict
            kwargs are forwarded to mpatches.Circle
        Returns
        -------
        patch : `~matplotlib.mpatches.Circle`
            Matplotlib patch object
        """

        import matplotlib.patches as mpatches

        val = self.pos.galactic
        center = (val.l.value, val.b.value)

        temp = dict(transform=ax.get_transform('galactic'),
                    radius=self.radius.value)
        kwargs.update(temp)
        patch = mpatches.Circle(center, **kwargs)

        return patch
```

So the WCSaxis is actually part of the API. My question is basically if I can copy&paste this function to astropy/regions or if we want to come up with another solution.

keflavich:
I understand.  For now, yes, go with that.  I'm not sure if there's a non-wcs-axis way to get the transforms easily.

astrofrog:
This looks great!

joleroi:
I changed the API and added a basic test

keflavich:
OK, will review.

keflavich:
@joleroi say if you think this is ready to merge, or merge it yourself if you'd like.  Others (@cdeil, @astrofrog, myself) are starting to hack on other aspects of the code.  It would be nice to have this merged, but would also be nice to have it merged with passing tests.

Issue #25, opened 2016-05-03
__contains__ scope?


joleroi:
At the moment the Region base classes have an abstract method:  ``__contains__``

https://github.com/astropy/regions/blob/master/regions/core/core.py#L195

The docstring suggests that this is supposed to work for a list of ``SkyCoords``

https://github.com/astropy/regions/blob/master/regions/core/core.py#L197

I don't think this works/is a good idea

```
In [35]: c1
Out[35]:
CircleSkyRegion
Center:<SkyCoord (Galactic): (l, b) in deg
    (0.0, 0.0)>
Radius:1.0 deg

In [36]: coords
Out[36]:
<SkyCoord (Galactic): (l, b) in deg
    [(1.2, 1.2), (0.5, 0.5), (0.7, 0.7)]>

In [37]: c1.__contains__(coords)
Out[37]: array([False,  True,  True], dtype=bool)

In [38]: coords in c1
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
/tmp/test.py in <module>()
----> 1 coords in c1

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

I looks like the ``in`` operator checks for a (non-array) boolean output of `` __contains__``. This makes sense if you think about statements like
```
if coord in regions:
    do something
``` 

I would suggest changing the API to ``regions.contains(coords)`` (without the underscores).

Do you agree @keflavich @astrofrog @bsipocz ?

astrofrog:
Sounds good, but we can leave ``__contains__`` for scalar cases?

joleroi:
cc @cdeil

cdeil:
For the record:

While writing the docs for region containment checks
https://github.com/astropy/regions/pull/50/files#diff-9bc8bf6e8d9db020cefce1fbc0426795R189
I had this same question, whether `points in region` should / could work for array-valued points.

There's some information here what `in` really does:
http://stackoverflow.com/questions/12244074/python-source-code-for-built-in-in-operator
http://stackoverflow.com/questions/37613288/in-statement-for-lists-tuples-of-floats

I used this example to play around:
```python
import numpy as np

class PixCoord:
    def __init__(self, x, y):
        self.x = x
        self.y = y
    
    def isscalar(self):
        return np.isscalar(self.x)

class PixelRegion:
    def __init__(self, center, radius):
        self.center = center
        self.radius = radius
    
    def __contains__(self, coord):
[...]

PR #26, opened 2016-05-03
Implement compound regions

joleroi:
* ``contains``

I don't know how to deal with ``area``, ``make_patch`` etc.

### DIFF
--- a/regions/core/compound.py
+++ b/regions/core/compound.py
@@ -29,16 +29,21 @@ class CompoundSkyRegion(SkyRegion):
     Represents the logical combination of two regions in sky coordinates.
     """
 
-    def __init__(self, region1, operator, region2):
+    def __init__(self, region1, region2, operator):
         self.region1 = region1
         self.region2 = region2
         self.operator = operator
 
-    def __contains__(self, skycoord):
-        raise NotImplementedError("")
+    def contains(self, skycoord):
+        return self.operator(self.region1.contains(skycoord),
+                             self.region2.contains(skycoord))
 
     def to_pixel(self, wcs, mode='local', tolerance=None):
         raise NotImplementedError("")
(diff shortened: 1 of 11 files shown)


joleroi:
This also changes the API as discussed in https://github.com/astropy/regions/issues/25

joleroi:
I would merge this and add further functionality in future PRs. Comments?

regions/core/compound.py
```diff
@@ -29,16 +29,21 @@ class CompoundSkyRegion(SkyRegion):
     Represents the logical combination of two regions in sky coordinates.
     """
 
-    def __init__(self, region1, operator, region2):
+    def __init__(self, region1, region2, operator):
```
astrofrog: The reason I had operator in the center is because it makes it clearer how to deal with non-symmetric cases (e.g. region 1 without region 2)

astrofrog:
@joleroi - that sounds fine. I have just one comment above at the moment. I think we will need some additional operators (for example the equivalent of ``-`` for sets).

astrofrog:
Indeed, ``area`` and ``make_patch``/``as_patch`` will be tricky for compound regions...

regions/core/compound.py
joleroi: There was a bug (https://github.com/astropy/regions/blob/master/regions/core/core.py#L67) . It was just easier to change this line instead of all the others .. Do you need the operator in the middle?

astrofrog: Ah I see - I personally think it would make it more readable, but up to you :)

joleroi: I'm more the RPN guy, you know ;)

astrofrog: 😆

PR #31, opened 2016-06-10
CircelPixelRegion.to_sky()

joleroi:
A round-trip test makes sure that this method is not totally crazy

### DIFF
--- a/regions/shapes/circle.py
+++ b/regions/shapes/circle.py
@@ -43,21 +43,21 @@ class CirclePixelRegion(PixelRegion):
     def to_shapely(self):
         return self.center.to_shapely().buffer(self.radius)
 
-    def to_sky(self, mywcs, mode='local', tolerance=None):
+    def to_sky(self, wcs, mode='local', tolerance=None):
         if mode != 'local':
-            raise NotImplementedError()
+            raise NotImplementedError
         if tolerance is not None:
-            raise NotImplementedError()
+            raise NotImplementedError
 
-        skypos = pixel_to_skycoord(self.center[0], self.center[1], mywcs)
-        xc, yc, scale, angle = skycoord_to_pixel_scale_angle(skypos, mywcs)
+        skypos = pixel_to_skycoord(self.center[0], self.center[1], wcs)
+        xc, yc, scale, angle = skycoord_to_pixel_scale_angle(skypos, wcs)
 
-        radius_sky = (self.radius / scale)
+        radius_sky = self.radius / scale
         return CircleSkyRegion(skypos, radius_sky)
 
     def to_mask(self, mode='center'):
         # TODO: needs to be implemented
-        raise NotImplementedError("")
+        raise NotImplementedError
 
     def as_patch(self, **kwargs):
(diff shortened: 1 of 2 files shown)


regions/shapes/circle.py
```diff
@@ -43,8 +44,16 @@ def to_shapely(self):
         return self.center.to_shapely().buffer(self.radius)
 
     def to_sky(self, mywcs, mode='local', tolerance=None):
-        # TODO: needs to be implemented
-        raise NotImplementedError("")
+        if mode != 'local':
+            raise NotImplementedError()
```
cdeil: You don't need to call the constructor, i.e. just always type `raise NotImplementedError` instead of `raise NotImplementedError()` and save two characters (which will be many characters over your lifetime)!

regions/shapes/circle.py
```diff
@@ -43,8 +44,16 @@ def to_shapely(self):
...
+        if tolerance is not None:
+            raise NotImplementedError()
+
+        skypos = pixel_to_skycoord(self.center[0], self.center[1], mywcs)
+        xc, yc, scale, angle = skycoord_to_pixel_scale_angle(skypos, mywcs)
+
+        radius_sky = (self.radius / scale)
```
cdeil: Remove parens here ... not needed.

regions/shapes/tests/test_circle.py
```diff
@@ -46,3 +46,18 @@ def test_plot():
...
+def test_transformation():
+    
+    skycoord = SkyCoord(3 * u.deg, 4 * u.deg, frame='galactic')
+    skycircle = CircleSkyRegion(skycoord, 2 * u.arcsec)
+    
+    headerfile = get_pkg_data_filename('data/example_header.fits')
+    h = getheader(headerfile)
+    wcs = WCS(h)
+    pixcircle = skycircle.to_pixel(wcs)
+    assert (pixcircle.center == (-50.5, 299.5)).all()
```
cdeil: Test is passing, but still: I never write exactly float compares, always `numpy.testing.asser_allclose`. Suggest you use that here.

regions/shapes/circle.py
```diff
@@ -43,8 +44,16 @@ def to_shapely(self):
         return self.center.to_shapely().buffer(self.radius)
 
     def to_sky(self, mywcs, mode='local', tolerance=None):
```
cdeil: `mywcs` is a weird argument name that's not used in other Astropy or affiliated package codes as far as I know.
Change to `wcs` possible here, no?

cdeil:
I left some minor inline comments.

regions/shapes/tests/test_circle.py
larrybradley: +1 for `assert_allclose`

regions/shapes/circle.py
larrybradley: +1

cdeil:
Looks good to me ... merging this now.

This PR doesn't implement all modes we want, and didn't get much review / feedback.
So everyone: follow-up issues and PRs welcome!

PR #42, opened 2016-06-23
Misc cleanup

cdeil:
This PR does some misc cleanup:
- [x] `regions/core/compound.py` with `'CompoundPixelRegion', 'CompoundSkyRegion'` is now part of the public API
- [x] `regions/core/pixcoord.py` with `PixCoord` is now part of the public API
- [x] `region_list_to_objects` is now part of the public API
- [x] Autoformat code
- [x] Remove some unused files
- [x] Clean up imports and add missing `__all__` lists

The diff looks horrible. I don't think this needs review, but if you want to have a look, go ahead.


cdeil:
I'm merging this now.
Comments and follow-up PRs are still welcome, of course!

PR #50, opened 2016-07-13
Write a first version of the getting started page

cdeil:
This pull request adds a first version of the high-level documentation for the `regions` package, on a "Getting started" page.

It also contains a bunch of code additions, fixes, cleanup and added tests.
I know it's bad practice to grow a PR this way, but I just started fixing things as I encountered something that didn't work properly.

If you disagree with something or notice further issues, please file a new issue and we'll fix it later.

PR #51, opened 2016-07-14
Implement PixCoord class

cdeil:
This PR implements the `PixCoord` class and adds tests.

I plan to add some more tests here for the to / from sky conversion.
High-level docs will be added later, in #50 .
Implementing and testing 2-dim or even n-dim PixCoord arrays will be done later.

Please review!

If I don't get feedback, I'll merge this tonight, because I want to finish up version 0.1.
Nothing will be set in stone, all feedback will still be welcome after 0.1.

### DIFF
--- a/regions/core/pixcoord.py
+++ b/regions/core/pixcoord.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 import numbers
 import numpy as np
 from astropy.coordinates import SkyCoord
+from .core import _DEFAULT_WCS_ORIGIN, _DEFAULT_WCS_MODE
 
 __all__ = ['PixCoord']
 
@@ -30,11 +31,6 @@ class PixCoord(object):
         Pixel coordinate x value
     y : float or array-like
         Pixel coordinate y value
-    origin : int
-        Origin of pixel coordinates (usually 0 or 1)
-    mode : {'all', 'wcs'}
-        Whether to do the transformation including distortions ('all')
-        or only including only the core WCS transformation ('wcs').
 
     Examples
     --------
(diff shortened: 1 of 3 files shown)


regions/core/pixcoord.py
```diff
@@ -1,26 +1,152 @@
...
-    """
+    # We define defaults origin and mode here for WCS transforms.
+    # These defaults are used in a few places as default keyword arguments to methods.
+    # The purpose is to ensure consistency across the codebase.
+    # They should not be modified by users, it's an implementation detail.
+    # TODO: is this the best place to put those defaults?
+    # Should we put them somewhere else?
+    # Or remove them and try to achieve consistency manually?
+    _DEFAULT_ORIGIN = 0
+    _DEFAULT_MODE = 'all'
```
keflavich: Maybe these should be module-level globals rather than class attributes?  That way, you won't be led to (incorrectly) think that modifying these values would have some effect in real code

cdeil: @keflavich - OK, I can make the default values class-level.

I think an even more important question is: should `origin` and `mode` be `PixCoord` data members?
Or should they not be data members, but only appear as options in the `to_sky` and `from_sky` (and other?) operations?
It certainly would be more efficient to just store `x` and `y`.
Concerning what users want, I'm not sure.
I don't see a use case where one would have to modify `origin` ... one advantage of the `PixCoord` class over just using `(x, y)` tuples is that it removes (or lessens) the possibility for off by one pixel errors.

Thoughts?

astrofrog: I personally would say that instead of having these as module/class level variables, they can be options that are specified during transformations.

cdeil: > I personally would say that instead of having these as module/class level variables, they can be options that are specified during transformations.

I agree. Unless others speak up in favour of keeping `origin` and / or `mode` as data members on the `PixCoord` class, I'll make this change.

cdeil:
I've implemented the feedback by @keflavich  and @astrofrog, I think this is ready for final review.

Please specifically check if this is OK:
- `_DEFAULT_WCS_ORIGIN=0` and `_DEFAULT_WCS_MODE='all'` are now global variables in `regions/core/core.py`
- `origin` and `mode` are now no longer data members of `PixCoord`, but taken as an option for methods that do a WCS transform.
- The tests. There is 100% test coverage, but overall the number of tests isn't large. If you think something is missing or tests should be structured in a better way, let me know now.

If I don't hear back I'll merge this tomorrow morning.
The next step is to use this class more in the regions classes and in the docs.

PR #53, opened 2016-07-18
Improve example dataset and docs

cdeil:
This pull request:

- [x] Adds a nice example dataset for docs and tests
- [x] Improves and extends the getting started page to use the new example dataset
- [x] Moves scripts to profile `regions` outside the package

At first I wanted to use the real `ExampleFermiDataset` and bundle a few 100 kB of example data in the repo.
But then I thought that this would be problematic when moving this to Astropy core, so I started to write code to simulate some example data.
I can still add the Fermi-LAT example dataset if others think that would be better.

This is work in progress ... I'll try to finish it tonight.


cdeil:
I'm merging this now. Will continue in other PRs.

PR #58, opened 2016-07-22
Add ds9_string_to_objects

cdeil:
Looking at
http://astropy-regions.readthedocs.io/en/latest/api.html#functions
I see `objects_to_ds9_string`, but no `ds9_string_to_objects`.

It seems that in the implementation (`ds9_parser(filename)`) the reading from file and parsing aren't separated.

This is weird, parsing should take string as input and then read should just read the file and call the parser with the string, no?
OK if I make this change now?

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -388,18 +388,73 @@ namely to buffer a rectangle, resulting in a polygon.
 ds9 region strings
 ==================
 
-TODO: add examples for `regions.write_ds9`, `regions.read_ds9`, `regions.objects_to_ds9_string`
-and `regions.region_list_to_objects`.
+The regions package provides functions to serialise and de-serialise Python lists of
+`regions.Region` objects to DS9 region strings: `~regions.ds9_objects_to_string`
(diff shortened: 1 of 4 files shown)


cdeil:
Commit attached here. @keflavich or anyone -- review?

keflavich:
:+1: I saw no problems.

cdeil:
I did expose the intermediate "region list" also in the high-level API and add some high-level docs.

They should appear here:
http://astropy-regions.readthedocs.io/en/latest/getting_started.html#ds9-region-strings

But they currently don't because RTD seems to error out for every build:
http://readthedocs.org/projects/astropy-regions/builds/4226037/
(the same happens for Gammapy today)

@astrofrog @keflavich It's not great, but maybe good enough for the 0.1 release?

In any case: test coverage, API and docs for the DS9 region part need work. Issues with suggestions or even pull requests welcome.

keflavich:
yeah, good enough.  If I'm lucky, I'll have some time to test and then maybe contribute again in two weeks...

Issue #55, opened 2016-07-21
Return a Mask class from the to_mask() method


larrybradley:
Just to follow-up on https://github.com/astropy/regions/issues/30, it would be better if the `to_mask()` method returns a `Mask` object.  The `Mask` object would be initialized with the mask array and the tuple slice and it could then handle the cases where the aperture is partially (or completely) outside of the image.

cdeil:
@larrybradley - Would you advocate starting a `Mask` class from scratch in `regions`? Or is there some class in `astropy` or other affiliated  packages that could be the starting point?

FYI: in Gammapy, we have been developing a [SkyMap](https://gammapy.readthedocs.io/en/latest/api/gammapy.image.SkyMap.html) class (will get renamed to `SkyImage` tomorrow). It has cut and paste via your Cutout class, but doesn't store offset as data member. I plan to integrate it with regions soon, but have no plan to propose it for Astropy core. We've discussed the possibility of adding an image class to Astropy several times, I think the conclusion was always that it's not welcome? Or maybe it was just that no-one made a PR?

I'm always +1 for more object-oriented API, grouping things like data and offset and making it convenient to interact with other objects like regions via methods instead of many standalone functions.

So @larrybradley - Could you elaborate a little what exactly you propose be added here, maybe put some pseudo-code what this `Mask` is supposed to do for the most common use cases?

larrybradley:
The `Mask` class was actually @astrofrog's idea and came out of a discussion I had with him earlier today (he was visiting STScI).  It would be a new class within `regions` (but it would likely reuse some code from `Cutout2D`).

The `to_mask()` method currently returns a mask array and tuple slice, but the region objects know nothing about the data on which the mask will be applied (e.g. performing photometry).  If the region falls partially or completely off the image, then the returned `tuple slice` will be incorrect.  Each and every time the mask is applied to data, the boundary conditions would need to be checked.  By returning a `Mask` object, the `Mask` object could then handle the boundary conditions (centralizing such code in a single place).  Perhaps something like

```
from regions import CirclePixelRegion
region = CirclePixelRegion(...)
mask = region.to_mask()     # returns a mask object
data_subset = mask.apply(data)
```
where the `apply()` method would handle the boundary checking.

Currently you would need to do:

```
from regions import CirclePixelRegion
region = CirclePixelRegion(...)
mask, slice_x, slice_y = region.to_mask()

# need code here to check slice boundary conditions

data_subset = mask * data[slice_y, slice_x]
```

[...]

PR #71, opened 2016-11-07
Implement to_mask for circular, elliptical, and rectangular regions

astrofrog:
This is a WIP that replaces/re-uses some of #17. Not at all ready for review/comments.


astrofrog:
@larrybradley - just FYI, I added a simple Mask class here I wrote during my flight, but just replace it with yours (I think ``regions.core.mask`` is the right place though)

larrybradley:
@astrofrog Did you allow for maintains to edit?  I tried to push to your branch, but it's not letting me.  I'll just submit a PR to your branch....

astrofrog:
@larrybradley - I had that issue before and realized it's because I'd set up a read-only remote. Anyway, I'll merge the PR, so no worries :) Thanks!

astrofrog:
@larrybradley - for Mask, I was thinking, maybe it's enough to have ``origin`` instead of having ``bbox_slice`` - the ``origin`` should be enough information to construct the slice, right?

astrofrog:
@larrybradley - see e.g. 50a57eb1c65ed4addd28c854fccb1156b8b2d863

larrybradley:
@astrofrog Yes, that works too.  For `photutils`, I simply passed the tuple of slices because I already have them defined for each aperture/position (e.g. see https://github.com/astropy/photutils/blob/master/photutils/aperture/circle.py#L148) instead of recalculating them.

So instead you would simply pass the `origin` as the equivalent of my`slice[0].start`, `slice[1].start`?

larrybradley:
Here's an example of the aperture `to_mask()` method from the `CircularMaskMixin` class:
https://github.com/astropy/photutils/blob/master/photutils/aperture/circle.py#L29

astrofrog:
> So instead you would simply pass the origin as the equivalent of myslice[0].start, slice[1].start?

Yes, exactly - the extent of the slice is redundant with the size of the mask (and were indeed checking for consistency before)

astrofrog:
Almost done... Need tests for Mask!

astrofrog:
This is ready for reviewing. What this does:

* Implements ``to_mask`` for circular, elliptical, and rectangular regions (the only missing mode is exact mode for rectangular regions)
* Adds reference tests using pytest-fits to compare the masks to reference versions stored in FITS files
* Adds a Mask class (fixes https://github.com/astropy/regions/issues/55)
* Remove 'any' and 'all' modes (fixes https://github.com/astropy/regions/issues/72)

This doesn't include tests for the Mask class yet, but @larrybradley told me he'd add them in a separate pull request.

Comments welcome!

cdeil:
@astrofrog - Thanks!

@larrybradley or @bsipocz - Do you have time to review this PR?

@astrofrog - What about high-level docs? Can you add them in this PR or do you prefer to make a reminder issue that this should be done in a follow-up PR?

E.g. this section on masks should be updated (currently points to photutils):
http://astropy-regions.readthedocs.io/en/latest/getting_started.html#masks

astrofrog:
@cdeil @bsipocz @larrybradley - I've added documentation, which you can also read here:

http://astropy-regions-astrofrog.readthedocs.io/en/latest/getting_started.html#masks

cdeil:
This docstring says that a mask and slices are returned:
http://astropy-regions-astrofrog.readthedocs.io/en/latest/api/regions.PixelRegion.html#regions.PixelRegion.to_mask

I think that's not the case, only a mask object is returned as shown in the examples:
http://astropy-regions-astrofrog.readthedocs.io/en/latest/getting_started.html#masks

Another thing I noticed is that the `to_mask` methods don't do input validation.
I.e. writing `mode='centre'` instead of `mode='center'` could error in a weird way or silently do something different from what the user wanted to do.
Maybe add input validation for such multiple-choice string options?
```
@staticmethod
def validate_mode(mode):
    if mode not in set('spam', 'ham', ...):
        raise ValueError(...)
```

cdeil:
The `Mask` docstring doesn't have any examples:
http://astropy-regions-astrofrog.readthedocs.io/en/latest/api/regions.Mask.html

Can you link from it to the high-level docs mask section, which has several examples?

The docs show `regions.Mask(mask, origin)`, but the docstring documents parameters `mask` and `bbox_slice`!?
Also, please add a one-line docstring for the `slices` attribute.
Maybe make the `__init__` arguments and data members of the class the same (if that makes sense, I didn't look closely).

astrofrog:
@cdeil - I think I've implemented all your comments (with help from @larrybradley)

cdeil:
@astrofrog - Thanks!

I had a quick look at the high-level docs with @adonath.

Some more feedback:

- The `array` and `slices` attributes should have a one-line description [here](http://astropy-regions-astrofrog.readthedocs.io/en/latest/api/regions.Mask.html)
- For `origin` I'd suggest you use a `PixCoord` object. The idea is that we use it everywhere, and not tuples in some places. The main advantage is that it's harder to screw up the x, y order. Suggesting to remove it is a valid proposal, but given that we have it at the moment, I think we should use it.
- If I understand correctly, the `pix_region.to_mask` method does two things: first compute minimal bounding box (including large image edge handling), then compute the mask array. My preference would be to introduce a simple small `BoundingBox` class (similar to `PixCoord`), that holds the two slices and again knows about axis order and has a nice `repr`. And maybe also expose the two computation steps separately, i.e. make it possible to just compute the bounding box for a given region, without computing the mask array?
- The high-level docs example never shows how the mask relates to the large original image. I would suggest ending the high-level docs section either with an example for that, or to end like this "this is a building block for aperture photometry. How to actually use it will be shows in `photutils` once that package has been updated to use `regions`". That's the plan, no?

Just FYI: What you've built here is a bit similar (but really pretty different) to what @adonath implemented for `gammapy.image.SkyImage`:
http://docs.gammapy.org/en/latest/image/sky_image.html#cutout-and-paste

cdeil:
So this copies over ~ 1000 lines of code from photutils, right?

When you merge it, could you please open a new issue in the regions or photutils tracker describing the next steps and plan to continue with this? I guess keep adding things to `regions` for a few months, and then in a year or so rewrite `photutils.aperture` to use that?

astrofrog:
@cdeil - thanks for the comments! Will work on this on the ✈️ tonight.

> So this copies over ~ 1000 lines of code from photutils, right?

Yes, the Cython code is copied over.

> When you merge it, could you please open a new issue in the regions or photutils tracker describing the next steps and plan to continue with this? I guess keep adding things to regions for a few months, and then in a year or so rewrite photutils.aperture to use that?

Will do - from talking to @larrybradley, that is indeed the plan.

astrofrog:
> The array and slices attributes should have a one-line description here

These are no longer public methods, so not needed

> For origin I'd suggest you use a PixCoord object. The idea is that we use it everywhere, and not tuples in some places. The main advantage is that it's harder to screw up the x, y order. Suggesting to remove it is a valid proposal, but given that we have it at the moment, I think we should use it.

Good idea - though as I started to implement it I realized that here we specifically want integer pixel coordinates, but PixCoord allows floating-point coordinates. Do we want to have a separate class for integer coordinates? If not, should we just check in Mask.__init__ that the PixCoord has integer values?

astrofrog:
It turns out that instead of using PixCoord for ``origin`` it is cleaner to use the BoundingBox class that I have now added to this PR.

astrofrog:
@cdeil @larrybradley - I've added an example of how to use the regions for simple photometry/statistics on images. However to do that I wanted to use load_star_image from photutils but it would be nice to not require photutils to build the docs. Would you have any objections to me moving the example photometry data from https://github.com/astropy/photutils-datasets/tree/master/data to data.astropy.org under a photometry directory? (to avoid too many repos with data)

cdeil:
@astrofrog - Thanks for all your work here. I'll have a look now.

> Good idea - though as I started to implement it I realized that here we specifically want integer pixel coordinates, but PixCoord allows floating-point coordinates. Do we want to have a separate class for integer coordinates? If not, should we just check in Mask.init that the PixCoord has integer values?

I'm not sure. The pro of a separate class would be to be more explicit and maybe better error messages to users. The con is a larger API and the extra implementation effort. Maybe a single class, but checking that dtype is int on input where indices are expected is the sweetspot?

@astrofrog @larrybradley  - What do you think?

> I've added an example of how to use the regions for simple photometry/statistics on images. However to do that I wanted to use load_star_image from photutils but it would be nice to not require photutils to build the docs. Would you have any objections to me moving the example photometry data from https://github.com/astropy/photutils-datasets/tree/master/data to data.astropy.org under a photometry directory? (to avoid too many repos with data)

Fine with me. Do whatever you think best.

---

There's an error in the docs build:
https://travis-ci.org/astropy/regions/jobs/175294628#L827

cdeil:
Here's the current [PixCoord](http://astropy-regions-astrofrog.readthedocs.io/en/latest/api/regions.PixCoord.html) and [BoundingBox](http://astropy-regions-astrofrog.readthedocs.io/en/latest/api/regions.BoundingBox.html). They are constructed as `PixCoord(x, y)` and `BoundingBox(jmin, jmax, imin, imax)`.

The problem I have with this is that I don't know how `i, j` related to `x, y`!?
The BoundingBox docstring doesn't give a clue.

My preference would be to use `BoundingBox(xmin, xmax, ymin, ymax)` to have the obvious correlation, and to explain to users that both `PixCoord` and `BoundingBox` can be used for integer pixel indices.

I think with `x` and `y` and some docs, we'll be able to be as clear as we can, whereas `i`, `j` I would never be able to remember which axis is which.

cdeil:
Concerning the tests added here and `pytest-fits`, I'm not sure.

Just to mention my concern: extra dev dependency, and once this is merged in `astropy` every user will have 100 extra small FITS files that get installed (there will be more files once we add more regions). In a PR like this, I can't verify on Github if the results are OK, I have to locally load the files in DS9, create corresponding regions to check ... would be about an hour of work.

The alternative would be to use very small reference images with just a few pixels (e.g. 3 x 2) and to store the reference results as text in the Python files, or to just assert on some pixels or on summary stats like sum.
This is the strategy we've done so far with gammapy.image
https://github.com/gammapy/gammapy/blob/master/gammapy/image/tests/test_core.py

@astrofrog @larrybradley - I'm fine with whatever you put here, and think will be easier to extend and maintain in `astropy.regions` for the coming years.

What is very nice is to have a good test reporting for large arrays about which elements were different like:
https://github.com/astrofrog/pytest-fits#test-failure-example
Maybe that could be moved to a utility function in Astropy?

docs/getting_started.rst
```diff
@@ -402,17 +402,198 @@ or `regions.PixelRegion.contains` methods:
...
+    plt.title("mode='subpixels', subpixels=20", size=9)
+    plt.imshow(mask4.data, cmap=plt.cm.viridis,
+               interpolation='nearest', origin='lower')
+
+As we've seen above, the :class:`~regions.Mask` objects have a ``data``
+attribute that contains a Numpy array with the mask values. However, if you
+have for example a circular region with a radius of 3 pixels at a pixel position
+of (1000, 1000), it would be inefficient to store a mask that has a size larger
+than this, so instead we store the mask using the minimal array that contains
+the mask, and the :class:`~regions.Mask` objects also include an ``origin``
```
cdeil: This needs to be updated, no?
You're saying from the high-level docs that mask has an origin attribute, but it now has a `bbox` attribute.
Also - please link to the `regions.BoundingBox` class at least once from the high-level docs.
And maybe link back from the class docstring to this section, or add one example there.

astrofrog:
@cdeil - for the testing - how about I change pytest-fits to allow multiple output formats, and instead of FITS I use say PNG and a text-based format (the former to see differences, the latter for more precision). Then we avoid FITS files, makes it easy to compare the results in GitHub PR, and we still have more accurate numerical tests and we preserve the nice test reporting for large arrays. Would that be an acceptable solution?

astrofrog:
Of course, I'd rename pytest-fits to pytest-raster for instance and it could have much wider applicability too that way. On the downside, we still have to have that plugin as a test-time dependency.

astrofrog:
However, since the plugin is an extra file, we could also simply bundle it in ``regions.extern`` for instance?

cdeil:
Yes, I think text-based reference files (YAML? JSON?) are better than FITS, they are more version control and pull request review friendly.

An option to have a single reference file instead of potentially 100s of files would be great.

Specifically for the mask tests here, you could also mention in the docstring in the test file how one could check the masks by overplotting the regions in ds9 or matplotlib.
I mean a check like https://github.com/astropy/regions/pull/74#issuecomment-259928714

For now, I would suggest to not bundle the pytest plugin here. Probably it'll evolve, maybe we'll start using it in Gammapy or other packages like photutils. We can still bundle it when it's more stable / when we merge this in Astropy core.
Maybe you could add a link to https://github.com/astrofrog/pytest-fits to the `regions` install docs?

astrofrog:
> An option to have a single reference file instead of potentially 100s of files would be great.

This isn't compatible with the idea of having PNG images that can be compared in GitHub PRs, unless I mosaic all the PNGs into a single PNG, but then any changes to the reference files will cause a large diff. Personally, I don't think it's a big issue to have <100 reference files (e.g. one PNG and one YAML file per test). If you like we can also not host these in the repo and instead host them on e.g. data.astropy.org and download on-the-fly as needed, like I do for WCSAxes. It's a little more work maintenance wise but does prevent the issue of adding many files to the repo/source install.

cdeil:
@astrofrog - What would be your preference to set this up and maintain it for `regions`. Do you want to commit the PNGs in the repo?

My first preference would be to only store a text file.
My second preference would be many text files.
My least favourite solution is many PNGs or FITS files.
(that said, I'm still OK with that last solution, if it's your preference)

In my experience what happens is this: the reference results are committed, and remain valid for a year. Then there's a code change or dependency change, and a bug report or pull request where this becomes relevant again. Usually what I do is go to the first pixel that has a difference and try to understand why it changed and if the old or new value is correct. I.e. I prefer to have the text file in the repo. Yes, having PNGs or FITS to quickly inspect things can sometimes be nice, but they can always be generated locally from the text files, I don't think they are useful enough to be stored in the repo.

astrofrog:
I agree text files would be the easiest, and best not to add PNG files to the repo too. Many files would be a lot easier than a single file though. I'll give it a try.

larrybradley:
>Would you have any objections to me moving the example photometry data from https://github.com/astropy/photutils-datasets/tree/master/data to data.astropy.org under a photometry directory? (to avoid too many repos with data)

Sounds fine with me.

fyi, I'm currently away on vacation -- I'll look at this (and several other issues) at the end of the week.

cdeil:
@astrofrog - Do you have time to finish this PR soon?

From my side, the only key point is that travis-ci passes.
I think we'll have changes / follow-up PRs in the coming months any ways, as people start using this for their work, and the best way to get good feedback is to put it in `master` or even to make a new release.

regions/core/bounding_box.py
```diff
@@ -0,0 +1,41 @@
...
+    @property
+    def slices(self):
+        """
+        The bounding box as a pair of `slice` objects.
+        """
+        return (slice(self.jmin, self.jmax),
+                slice(self.imin, self.imax))
+
+    @property
+    def extent(self):
```
cdeil: Maybe add a bbox method or attribute that gives a rectangle for quick MPL plotting (or dumping to DS9 region).
This is often useful when debugging code or algorithms using bounding boxes.

astrofrog:
> My preference would be to use BoundingBox(xmin, xmax, ymin, ymax) to have the obvious correlation, and to explain to users that both PixCoord and BoundingBox can be used for integer pixel indices.

Done (using ixmin/ixmax and iymin/iymax to highlight these are integers)

astrofrog:
Example of bounding box plotting:

```python
import numpy as np

center = PixCoord(6.6, 7.2)
reg = CirclePixelRegion(center, 5.2)
mask = reg.to_mask(mode='center')

fig = plt.figure()
ax = fig.add_subplot(1,1,1)
ax.imshow(np.random.random((20,20)), interpolation='nearest', cmap=plt.cm.viridis)
ax.add_patch(reg.as_patch(facecolor='none', edgecolor='white'))
ax.add_patch(mask.bbox.as_patch(facecolor='none', edgecolor='white'))
```

bbox

cdeil:
> Example of bounding box plotting

**Very** nice illustration!
Add to high-level docs or BoundingBox docstring?

cdeil:
For plotting, can we always put y=0 at the bottom, i.e. show images how ds9 displays them?

astrofrog:
@cdeil - will do. Just about to add the photutils data to data.astropy.org to avoid the photutils dependency here.

astrofrog:
@cdeil - I think this is good to go. Do you want me to purge the reference FITS files from the git history though? I don't think they are very big, but just let me know what you think.

cdeil:
> Do you want me to purge the reference FITS files from the git history though?

If it's quick and simple: yes.
If it's tricky: no.

I don't care about git history much, but maybe you or others in Astropy do, in case you want to preserve this history when merging in the core repo (which I also don't care about, OK to just make a copy and note in the changelog who contributed here).

cdeil:
I have to run now. Please merge, and I'll read the docs and try it out in the coming days.
Might come back with further comments or make follow-up pull requests myself then.

astrofrog:
@cdeil - I cleaned the FITS files out of the history, so I'll go ahead and merge now!

PR #77, opened 2016-11-21
Fix plotting of SkyCircleRegion

### LINKED ISSUE #76: SkyCircleRegion matplotlib plot issue
In #74 @joleroi added [docs/plot_compound.py](https://github.com/astropy/regions/blob/master/docs/plot_compound.py) which generates this plot:
http://astropy-regions.readthedocs.io/en/latest//plot_compound.png

As discussed in #74, there is an issue, the region is not drawn correctly.
To @joleroi and me the script looks OK, so we don't know what's wrong.

@astrofrog or @larrybradley or anyone - if you could have a look, that would be great!

astrofrog:
This fixes #76. See https://github.com/astrofrog/wcsaxes/issues/185 for some background about this and http://wcsaxes.readthedocs.io/en/latest/overlays.html#spherical-patches for documentation.

WCSAxes doesn't have classes for other shapes though, so that would need to be implemented before fixing those. But fundamentally the issue is that we can't simply use the built-in Matplotlib patch classes and instead have to define high-resolution polygons that go through the right coordinates.

test

### DIFF
--- a/regions/shapes/circle.py
+++ b/regions/shapes/circle.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 import math
 from astropy.coordinates import Angle
 from astropy.wcs.utils import pixel_to_skycoord
+from astropy import units as u
 from ..core import PixelRegion, SkyRegion
 from ..utils.wcs_helpers import skycoord_to_pixel_scale_angle
 
@@ -66,6 +67,7 @@ class CirclePixelRegion(PixelRegion):
         raise NotImplementedError
 
     def as_patch(self, **kwargs):
+
         import matplotlib.patches as mpatches
 
         xy = self.center.x, self.center.y
@@ -139,16 +141,16 @@ class CircleSkyRegion(SkyRegion):
         return CirclePixelRegion(center, radius)
 
     def as_patch(self, ax, **kwargs):
-        import matplotlib.patches as mpatches
 
-        val = self.center.icrs
-        center = (val.ra.to('deg').value, val.dec.to('deg').value)
+        from wcsaxes.patches import SphericalCircle
 
-        temp = dict(transform=ax.get_transform('icrs'),
-                    radius=self.radius.to('deg').value)
-        kwargs.update(temp)
+        c_icrs = self.center.icrs
+
         for key, value in self.visual.items():
             kwargs.setdefault(key, value)
-        patch = mpatches.Circle(center, **kwargs)
+
+        patch = SphericalCircle((c_icrs.ra, c_icrs.dec), self.radius,
+                                vertex_unit=u.degree,
+                                transform=ax.get_transform('icrs'), **kwargs)
 
         return patch


astrofrog:
Another thought - I think ideally WCSAxes should actually be using regions' to_pixel to do the conversion of the patch from sky to pixel coordinates - so there is a bit of circularity here. Maybe to_patch for sky regions should actually just call to_pixel and then return a patch in pixel coordinates.

cdeil:
> WCSAxes doesn't have classes for other shapes though, so that would need to be implemented before fixing those. But fundamentally the issue is that we can't simply use the built-in Matplotlib patch classes and instead have to define high-resolution polygons that go through the right coordinates.

Good to know.

> Maybe to_patch for sky regions should actually just call to_pixel and then return a patch in pixel coordinates.

Sounds good to me.

@astrofrog - What do you recommend exactly?

I guess there's a few options?

1. sky region -> pixel region -> mpl shape -> draw
2. sky region -> pixel region -> pixel polygon -> mpl polygon -> draw
3. sky region -> sky polygon -> pixel polygon -> mpl polygon -> draw

Intuitively I'd think  doing 3. for any sky region is possible and maybe a good way to go? Is it a concern / goal that we want to offer the option to draw regions identically to how ds9 does it?
There's maybe an issue with how to expose an option that controls the number of points, or how to polygonise shapes well?

Concerning the question of whether MPL code lives in `astropy.regions` or `wcsaxes`, I have no opinion at this point. We could also take all MPL code out here and leave it to `wcsaxes` to draw? Are there significant pros / cons either way? Or will it be roughly the same for users and maintainers either way?

@astrofrog - Can you make a decision here, or does this need more discussion / feedback?
Can you outline the tasks that need to be done, so that others (including me) can help implement this in the near future?

cdeil:
There's a build fail:
https://travis-ci.org/astropy/regions/jobs/177697553#L726
But I guess we don't want to merge this, and instead go for a solution that will work for all regions some day?

astrofrog:
> sky region -> pixel region -> mpl shape -> draw

This would be my preference. See example lower down.

> There's maybe an issue with how to expose an option that controls the number of points, or how to polygonise shapes well?

The plan was that this was a kwarg in the ``to_pixel`` method for sky regions (see ``tolerance`` [here](https://github.com/astropy/regions/blob/master/regions/core/core.py#L270))

So just to clarify, this is how it could work:

```python
>>> reg = SkyCircleRegion(...)
>>> preg = reg.to_pixel(mode='local')  # returns a PixelCircleRegion
>>> ax.add_patch(preg)  # adds a Circle patch in pixel space
```

```python
>>> reg = SkyCircleRegion(...)
>>> preg = reg.to_pixel(mode='affine')  # returns an PixelEllipseRegion
>>> ax.add_patch(preg)  # adds an Ellipse patch in pixel space
```

```python
>>> reg = SkyCircleRegion(...)
>>> preg = reg.to_pixel(mode='full')  # returns an PixelPolygonRegion
>>> ax.add_patch(preg)  # adds a Polygon patch in pixel space
```

In the latter example, ``to_pixel`` can optionally take a tolerance.

One could optionally then just define ``as_patch`` for sky regions to be:

```python
def as_patch(self, mode='local', tolerance=0, **kwargs):
    pixel_reg = self.to_pixel(mode=mode, tolerance=tolerance)
    return pixel_reg.as_patch(**kwargs)
```

which would simplify the code above while making sure the sky -> pixel conversion is only defined once, and we never have to rely on fancy WCSAxes classes.

cdeil:
@astrofrog - Note that region pixel / sky conversion always needs a `wcs` as input:
http://astropy-regions.readthedocs.io/en/latest/getting_started.html#transformations
So that would be an extra argument in the code examples from your last comment, right?

@astrofrog - So basically you're proposing to get rid of direct matplotlib plotting of all **sky** regions (i.e. `as_mpl_patch` or `plot_mpl` methods). Let the user convert to pixel regions (or other packages like `wcsaxes`). Right?

Sounds good to me. It does let us concentrate on regions and functionality now, instead of MPL transform and MPL sky region plotting API questions.
Putting no sky region MPL plotting now doesn't mean we can't add it later if we think it belongs in `astropy.regions`.

@larrybradley @keflavich @adonath @joleroi - Please comment on this question!

astrofrog:
> So basically you're proposing to get rid of direct matplotlib plotting of all sky regions (i.e. as_mpl_patch or plot_mpl methods). Let the user convert to pixel regions (or other packages like wcsaxes). Right?

Yes, or we can just add the simple 3-line function I was mentioning that adds that convenience for the user, but no more.

adonath:
:+1: To remove the `.as_patch()` methods and later add e.g. a `WCSAxes.add_region()` method.

larrybradley:
@astrofrog In your examples above I think you are missing the `as_patch()` method on the pixel region when using `ax_add_patch()`, e.g.

```
>>> reg = SkyCircleRegion(...)
>>> preg = reg.to_pixel(mode='local')  # returns a PixelCircleRegion
>>> ax.add_patch(preg.as_patch())  # adds a Circle patch in pixel space
```

larrybradley:
I think we should keep @astrofrog's simple 3-line `as_patch()` method on sky regions that automatically converts to pixel regions before creating the patch.  Is there any reason not to use that?

cdeil:
I've started a new PR in #85.
The plan is that we close this one without merging once #85 is in.

cdeil:
> I think we should keep @astrofrog's simple 3-line as_patch() method on sky regions that automatically converts to pixel regions before creating the patch. Is there any reason not to use that?

Would that be the correct way to do it?
```python
def as_patch(self, wcs, mode='local', tolerance=0, **kwargs):
    pixel_reg = self.to_pixel(wcs=wcs, mode=mode, tolerance=tolerance)
    return pixel_reg.as_patch(**kwargs)
```
And that would be on the `SkyRegion` base class and work for all sky regions?

I'm fine to keep it as a convenience. But I don't have much time this week, at the moment we have no tests at all. Could also be added later. Or help welcome. (We can continue with #85 or throw it away if someone else makes another PR.)

I'm offline now, will continue with this tomorrow.

larrybradley:
@cdeil I think that would work on the `SkyRegion` base class, assuming `to_pixel` is an abstract method of `SkyRegion` (i.e. all sky regions have that method).

cdeil:
@larrybradley - I've thought about this again, and prefer not to add the two-line `SkyRegion.as_patch` method in #85. Mainly because I don't have time to add tests and docs and to look at the exact shapes that come out. To a minor degree because I'm not sure it's useful.

Compare:
```python
patch = sky_region.to_pixel(wcs, mode, tolerance).as_patch(**kwargs)
```
to
```python
patch = sky_region.as_patch(wcs, mode, tolerance, **kwargs)
```

I actually prefer the first version, even if it's a few extra characters to type, because it means the parameters that define sky -> pix transformation are separated from MPL plotting options.

Anyways, if someone has time to follow-up my #85 in the coming days with a PR adding tests and better docs (and maybe also extra methods) for sky image plotting, that would be great!
If no, it'll be in the 0.3 release.

larrybradley:
@cdeil :+1: to remove the `SkyRegion.as_patch`.

cdeil:
I just now merged #85, which fixes this issue.

I don't think we want to ever merge this PR, but it contains useful discussion concerning plotting of sky regions. So I'm keeping it open for now, and I'm moving it to the 0.3 milestone.

We should create (and add tests for asserting on polygon vertex positions) the functionality that's now in `wcsaxes.SphericalCircle`, and generally try to make `astropy.regions` work in an as-simple and smooth way together with MPL and WCSAxes.

cdeil:
@astrofrog - I think this code was removed, no? Close this old PR? Or is there something that still eventually can be merged here?

astrofrog:
Yep, this is no longer relevant

PR #32, opened 2016-06-10
Add first version of API docs

cdeil:
At PyAstro16 @joleroi set up online docs for this regions repo here:
http://astropy-regions.readthedocs.io/en/latest/

This pull request adds a first version of the API docs.

There's a few warnings in the API docs build like this that I don't know how to fix.
@keflavich @astrofrog or @bsipocz - Do you know how to address those? If yes, can you make a PR against this PR?
```
<autosummary>:None: WARNING: toctree contains reference to nonexisting document 'api/regions.shapes.PixelRegion'
<autosummary>:None: WARNING: toctree contains reference to nonexisting document 'api/regions.shapes.SkyRegion'
None:None: WARNING: toctree contains reference to nonexisting document 'api/regions.shapes.PixelRegion'
None:None: WARNING: toctree contains reference to nonexisting document 'api/regions.shapes.SkyRegion'
```
https://gist.github.com/cdeil/539490416ffee012f3a1081b90940168

If I don't hear back I might merge this soon, because something with warnings is better than nothing. :-)

PR #33, opened 2016-11-28
Is regions.utils private?

cdeil:
While adding the API docs in #32 I noticed that there's a `regions.utils`:
https://github.com/astropy/regions/tree/master/regions/utils

@keflavich @astrofrog - Is this still needed or is that functionality now available in Astropy core?
If we keep it, should it be part of the public API?
If no, can we rename `utils` to `_utils` to make it clear it's private?


cdeil:
I've attached a commit that does the rename `regions.utils` -> `regions._utils` here.
I'll merge if travis-ci passes, so if someone wants to have a look, do it now.

PR #81, opened 2016-11-28
Round to int, not float

cdeil:
@astrofrog added bounding box methods that call `round`:
https://github.com/astropy/regions/search?utf8=%E2%9C%93&q=round

Now here's what I think is an amazing catch by PyCharm (or a false positive if I misunderstand the situation): It points out that e.g. `CirclePixelRegion.bounding_box` calls `BoundingBox(ixmin, ixmax, iymin, iymax)` with floats, but it should call it with ints.

Now first of all, I think it knows about the expected types because it's parsing the `BoundingBox` Numpy-format docstring. Also, it asked me if this project should be simultaneously Python 2.7 and 3 compatible and I said yes. And lastly, on Python 2, `round` returns float:
```
$ python2
>>> type(round(0.3))
<type 'float'>

$ python3
>>> type(round(0.3))
<class 'int'>
```

@astrofrog @adonath - Is this indeed a potential issue? I think we should change callers to always call `int`, instead of auto-casting of ints stored in `float` in `BoundingBox.__init__`? Should we add an extra check that ints are passed in? What's the best (correct, short) way to write this, i.e. what code changes should we make?

Should we do a general review of `round` code in `astropy`, `photutils`, `gammapy` or just make a change here? Does one of you have time to make a PR or should I?

### DIFF
--- a/regions/core/bounding_box.py
+++ b/regions/core/bounding_box.py
@@ -48,6 +48,32 @@ class BoundingBox(object):
         self.iymin = iymin
         self.iymax = iymax
 
+    @classmethod
+    def _from_float(cls, xmin, xmax, ymin, ymax):
+        """Smallest BoundingBox that fully contains a given float rectangle.
+
+        Parameters
+        ----------
+        xmin, xmax, ymin, ymax : float
+            Rectangle with arbitrary float coordinates
+
+        Returns
+        -------
+        bbox : `BoundingBox`
+            Smallest bounding box fully containing the rectangle
+        """
+        # We use round here because if the region extends
+        # to e.g. -0.4, it's enough to set the bounding box lower value to 0
+        # because the 0-th pixel goes from -0.5 to 0.5. At the upper end we add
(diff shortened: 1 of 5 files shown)


cdeil:
@astrofrog - I've attached a commit here, please have a look.

For now, I've made `BoundingBox._from_float` private, and mainly added it to avoid the triple code duplication and fix the `int(round())` issue in one place.

If someone has a suggestion for name / options / semantics, I could implement that here and we could make it public.

In any case, I plan to add a regression test here and maybe also the `if not isinstance(ixmin, numbers.Integral)` input validation.

cdeil:
I've added a better docstring and tests.

Concerning int type validation, I'll file a separate issue, I don't want to do it in this PR.

larrybradley:
I was just reviewing this.  I'll submit a new PR.

larrybradley:
For the record, `photutils` does not use `np.round` anywhere (and that's a good thing!).

PR #84, opened 2016-11-28
Fix bounding box "rounding"

larrybradley:
This PR is a followup to #81, which used the `round` function.  We don't want to use `round` to compute the bounding-box edges because of its behavior at half-pixel values.  For such cases `round` (in Python 3) rounds to the nearest even number, e.g. `round(0.5) = 0.0` but `round(1.5) = 2.0`.  That behavior may make sense in a statistical sense, but it's incorrect here for computing bounding box edges.

In this PR, I use what I implemented in my refactor of `photutils.aperture`:  
```
imin = np.floor(value + 0.5).astype(int)
imax = np.floor(value + 1.5).astype(int)
```

which "rounds" consistently for the bounding box edges.

Also note that the above behavior changed for Python 3.  In Python 2, `round` would always round away from zero, e.g. `round(0.5) = 1.0` and `round(1.5) = 2.0`.  This is yet another inconsistency that we want to avoid (and another reason why `round` should not be used).

### DIFF
--- a/regions/core/bounding_box.py
+++ b/regions/core/bounding_box.py
@@ -54,36 +54,42 @@ class BoundingBox(object):
 
     @classmethod
     def _from_float(cls, xmin, xmax, ymin, ymax):
-        """Smallest BoundingBox that fully contains a given float rectangle.
+        """
+        Return the smallest bounding box that fully contains a given
+        rectangle defined by float coordinate values.
 
-        Following the pixel convention, the pixel edges correspond to these
-        pixel coordinates for example for three pixels:
+        Following the pixel index convention, an integer index
+        corresponds to the center of of a pixel and the pixel edges span
+        from (index - 0.5) to (index + 0.5).  For example, the pixels
+        edges of these three pixels are:
 
-        - pixel 0 from -0.5 to 0.5
-        - pixel 1 from 0.5 to 1.5
-        - pixel 2 from 1.5 to 2.5
+        - pixel 0: from -0.5 to 0.5
+        - pixel 1: from 0.5 to 1.5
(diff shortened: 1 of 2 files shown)


cdeil:
@larrybradley - Thanks!

You put this:
```
ixmin = np.floor(xmin + 0.5).astype(int)
```
Is this the same?
```
ixmin = int(xmin + 0.5)
```
If yes, I think I would slightly prefer it.
But in any case, just a thought ... we'll leave what you think best.

Another question: should we really have a test for `0.5`?
It's a fractional float. Is it really guaranteed to round the way you have in the test on all machines?

larrybradley:
>You put this:

>ixmin = np.floor(xmin + 0.5).astype(int)
>Is this the same?
>
>ixmin = int(xmin + 0.5)
>If yes, I think I would slightly prefer it.
>But in any case, just a thought ... we'll leave what you think best.

For non-negative `xmin` they are the same, but they can differ for negative `xmin`, e.g.

```
>>> xmin = -1
>>> np.floor(xmin + 0.5).astype(int)
-1

>>> int(xmin + 0.5)
0
```

Negative values in the bounding boxes can occur if the region is only partially on the image.  The `Mask` methods handle those cases when applied to data.

larrybradley:
>Another question: should we really have a test for 0.5?
>It's a fractional float. Is it really guaranteed to round the way you have in the test on all machines?

I'm not actually sure.

cdeil:
> I'm not actually sure.

+1 to remove that test and then merge.

larrybradley:
@cdeil Done!

larrybradley:
@cdeil I removed the 0.5 rounding example in `master`.

PR #85, opened 2016-11-28
Improve shape classes and MPL plotting

cdeil:
This PR:

- [x] Removes sky region `as_patch` and `plot` as discussed in #76 and #77 
- [x] Group XYZRegion class tests in TestXYZRegion classes
- [x] Implement `EllipsePixelRegion.as_patch` (and add example to docstring)
- [x] Implement `RectanglePixelRegion.as_patch` (and add example to docstring)
- [x] Change to major / minor argument order in ellipse to be consistent with MPL / ds9 / photutils
- [x] Change to width / height argument order in rectangle to be consistent with MPL / ds9 / photutils
- [x] Add tests for pixel region `as_patch` methods
- [x] Updates the region MPL plot examples in the docs (to address #76)
- [x] Updates docs matplotlib plotting section with more info / examples

### DIFF
--- a/regions/shapes/ellipse.py
+++ b/regions/shapes/ellipse.py
@@ -133,21 +133,21 @@ class EllipseSkyRegion(SkyRegion):
     ----------
     center : `~astropy.coordinates.SkyCoord`
         Center position
-    minor : `~astropy.units.Quantity`
-        Minor radius
     major : `~astropy.units.Quantity`
         Major radius
+    minor : `~astropy.units.Quantity`
+        Minor radius
     angle : `~astropy.units.Quantity`
         The rotation angle of the ellipse.
         If set to zero (the default), the major
         axis is lined up with the longitude axis of the celestial coordinates.
     """
 
-    def __init__(self, center, minor, major, angle=0. * u.deg, meta=None, visual=None):
+    def __init__(self, center, major, minor, angle=0. * u.deg, meta=None, visual=None):
         # TODO: use quantity_input to check that height, width, and angle are angles
         self.center = center
-        self.minor = minor
         self.major = major
+        self.minor = minor
         self.angle = angle
         self.meta = meta or {}
         self.visual = visual or {}
@@ -156,11 +156,11 @@ class EllipseSkyRegion(SkyRegion):
         data = dict(
             name=self.__class__.__name__,
             center=self.center,
-            minor=self.minor,
             major=self.major,
+            minor=self.minor,
             angle=self.angle,
         )
-        fmt = '{name}\ncenter: {center}\nminor: {minor}\nmajor: {major}\nangle: {angle}'
+        fmt = '{name}\ncenter: {center}\nmajor: {major}\nminor: {minor}\nangle: {angle}'
         return fmt.format(**data)
 
     @property
--- a/regions/shapes/ellipse.py
+++ b/regions/shapes/ellipse.py
@@ -24,6 +24,32 @@ class EllipsePixelRegion(PixelRegion):
         The rotation angle of the ellipse.
         If set to zero (the default), the major
         axis is lined up with the x axis.
+
+    Examples
+    --------
+
+    .. plot::
+        :include-source:
+
+        import numpy as np
+        from astropy.modeling.models import Ellipse2D
+        from astropy.coordinates import Angle
+        from regions import PixCoord, EllipsePixelRegion
+        import matplotlib.pyplot as plt
+        x0, y0 = 15, 10
+        a, b = 8, 5
+        theta = Angle(30, 'deg')
+        e = Ellipse2D(amplitude=100., x_0=x0, y_0=y0, a=a, b=b, theta=theta.radian)
+        y, x = np.mgrid[0:20, 0:30]
+        fig, ax = plt.subplots(1, 1)
+        ax.imshow(e(x, y), origin='lower', interpolation='none', cmap='Greys_r')
+
+        center = PixCoord(x=x0, y=y0)
+        reg = EllipsePixelRegion(center=center, major=a, minor=b, angle=theta)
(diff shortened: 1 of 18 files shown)


regions/shapes/ellipse.py
```diff
@@ -102,16 +102,27 @@ def to_mask(self, mode='center', subpixels=5):
...
     def as_patch(self, **kwargs):
-        # TODO: needs to be implemented
-        raise NotImplementedError
+        """Matplotlib patch object for this region (`matplotlib.patches.Ellipse`).
+        """
+        from matplotlib.patches import Ellipse
+        xy = self.center.x, self.center.y
+        width = 2 * self.major
+        height = 2 * self.minor
+        # TODO: think about how to map major / minor / angle to MPL parameters.
```
cdeil: @larrybradley @astrofrog - OK if I change the argument order for ellipse to major / minor?
This seems more natural to me and is more in line with http://photutils.readthedocs.io/en/latest/api/photutils.EllipticalAperture.html .

What about the `angle` passed to MPL.
Does anyone know if it's OK as defined here?
I was looking at this for guidance:
http://photutils.readthedocs.io/en/latest/_modules/photutils/aperture/ellipse.html#EllipticalAperture.plot

cdeil: Same for rectangle ... I'll go ahead and follow the lead of MPL and photutils and change to width/height argument order:
http://photutils.readthedocs.io/en/latest/api/photutils.RectangularAperture.html
Let me know if you disagree.

regions/shapes/tests/test_rectangle.py
```diff
@@ -1,26 +1,52 @@
...
+        expected = 'RectanglePixelRegion\ncenter: PixCoord(x=3, y=4)\nwidth: 4\nheight: 3\nangle: 5.0 deg'
+        assert str(self.reg) == expected
 
+    @pytest.mark.skipif('not HAS_MATPLOTLIB')
+    def test_as_patch(self):
+        patch = self.reg.as_patch()
+        assert_allclose(patch.xy, (3, 4))
+        assert_allclose(patch.get_width(), 4)
+        assert_allclose(patch.get_height(), 3)
+        assert_allclose(patch._angle, 5)
```
cdeil: @astrofrog @larrybradley - I couldn't find a public attribute or method to get at the `angle` in a `matplotlib.patches.Rectangle`. Does it exist!?

cdeil:
@astrofrog - Is this a bug in `pytest_arraydiff`?
https://travis-ci.org/astropy/regions/jobs/179802570#L1056
Or am I doing something wrong?
I don't see this locally, must be a different version?

cdeil:
@astrofrog @larrybradley - This is ready for final review.

regions/shapes/tests/test_rectangle.py
larrybradley: @cdeil I looked in the `matplotlib` source code and there's no public attribute or method (that's odd), only the `_angle` attribute.

regions/shapes/ellipse.py
larrybradley: I'm not sure where you changed the major/minor order, but what you have here is correct for the ellipse.  Photutils defines the `theta` as the counterclockwise rotation of the semimajor axis from the `x` axis.  This is the same for the `EllipsePixelRegion`.

https://github.com/cdeil/regions/blob/387c5a60a699bffdf3546c620cf6396fd5a9d06c/regions/shapes/ellipse.py#L25

larrybradley: I don't see any changes to rectangle.  `photutils.RectangularAperture` is correct as defined, where the `width` side is along the `x` axis for `theta=0` and `theta` increases counterclockwise.

Note that the all of these plots require the "normal" display convention where the `(0, 0)` pixel is at the lower left (i.e. `origin='lower'`).  Otherwise the rotation angle will be in the opposite direction.

cdeil:
The `np.savetxt` fails with `fmt` are a known numpy bug:
https://github.com/numpy/numpy/pull/4053#issuecomment-263808221
I'll change to bytes literal now.
@astrofrog - Might even be worth converting `fmt` to bytes in `pytest-arraydiff` if users pass `bytes`, to spare other users the trouble of running into this issue.

cdeil:
Concerning `matplotlib.patches.Rectangle.angle`, I've filed this feature request:
https://github.com/matplotlib/matplotlib/issues/7536
For now, I'll uncomment that assert on `_angle` and leave a TODO comment and reference to that MPL issue.

regions/shapes/ellipse.py
cdeil: @larrybradley - Thank you. I've fixed added two more commits here that fix up the rectangle as_patch and added a plot to the docstring for rectangle and ellipse. Docstring updates are deferred to #86

joleroi:
I lost track of all the plotting discussion, however the current state of the compound regions example plot does not look satisfactory at all, does it?

http://astropy-regions.readthedocs.io/en/latest/getting_started.html#compounds

As far as I understand, @cdeil does not want to merge https://github.com/astropy/regions/pull/77. 

> I just now merged #85, which fixes this issue.

> I don't think we want to ever merge this PR, but it contains useful discussion concerning plotting of sky regions. So I'm keeping it open for now, and I'm moving it to the 0.3 milestone.

So how are we going to improve the plotting issue?

cdeil:
@joleroi and I discussed this over lunch. The plot isn't incorrect, it just plots pixel regions.

I agree it's confusing for illustrating compounds, the points and regions should be consistently in pixels or in sky, or there could even be one example for each.

Also, there should be an illustration of the differences between pixel and sky regions further up in the docs.

I don't have time, and I think we should start by implementing our polygon classes, and only then come back to docs / testing for sky regions.

But like I said in #63, if someone has time to make a pull request to improve `regions` this week, great! If not, we'll try to do it in the coming months.

PR #90, opened 2016-11-30
ds9 reader corrections/improvements

keflavich:
Mostly minor:

 * warn if a region is skipped when reading
 * read 'box' regions (we accepted only 'rectangle' before)
 * allow for whitespace after commas

### DIFF
--- a/regions/io/read_ds9.py
+++ b/regions/io/read_ds9.py
@@ -26,13 +26,18 @@ class DS9RegionParserWarning(AstropyUserWarning):
     warnings
     """
 
-def read_ds9(filename):
+def read_ds9(filename, errors='strict'):
     """Read a ds9 region file in as a list of region objects.
 
     Parameters
     ----------
     filename : str
         The file path
+    errors : ``warn``, ``ignore``, ``strict``
+      The error handling scheme to use for handling parsing errors.
+      The default is 'strict', which will raise a ``ValueError``.
+      ``warn`` will raise a warning, and ``ignore`` will do nothing
+      (i.e., be silent).
 
     Returns
     -------
@@ -42,7 +47,7 @@ def read_ds9(filename):
     with open(filename) as fh:
         region_string = fh.read()
 
-    return ds9_string_to_objects(region_string)
+    return ds9_string_to_objects(region_string, errors=errors)
 
 
 def parse_coordinate(string_rep, unit):
@@ -273,6 +278,8 @@ def ds9_string_to_region_list(region_string, errors='strict'):
     regions = []
     composite_region = None
 
+    global_meta = {}
+
     # ds9 regions can be split on \n or ;
     lines = []
     for line_ in region_string.split('\n'):
@@ -281,8 +288,12 @@ def ds9_string_to_region_list(region_string, errors='strict'):
             parsed = line_parser(line, coordsys, errors=errors)
             if parsed in coordinate_systems:
                 coordsys = parsed
+            elif parsed and (parsed[0] == 'global'):
+                # set some global metadata from the 'global' header
+                _, global_meta = parsed
             elif parsed:
                 region_type, coordlist, meta, composite, include = parsed
+                meta.update(global_meta)
                 meta['include'] = include
                 log.debug("Region type = {0}.  Composite={1}"
                           .format(region_type, composite))
@@ -330,6 +341,14 @@ def line_parser(line, coordsys=None, errors='strict'):
     """
     if errors not in ('strict','ignore','warn'):
         raise ValueError("``errors`` must be one of strict, ignore, or warn")
+
+    if '# Region file format' in line and line[0] == '#':
+        # This is just a file format line, we can safely skip it
+        return
+
+    # special case / header: parse global parameters into metadata
+    if line.lstrip()[:6] == 'global':
+        return global_parser(line)
 
     region_type_search = region_type_or_coordsys_re.search(line)
     if region_type_search:
@@ -482,3 +501,6 @@ def meta_parser(meta_str):
               for ii, jj in zip(equals_inds, equals_inds[1:] + [None])}
 
     return result
+
+def global_parser(line):
+    return "global", meta_parser(line)
--- a/regions/io/read_ds9.py
+++ b/regions/io/read_ds9.py
@@ -177,7 +177,7 @@ def ds9_region_list_to_objects(region_list):
                 reg = polygon.PolygonPixelRegion(coord_list[0])
             else:
                 raise ValueError("No central coordinate")
-        elif region_type == 'rectangle':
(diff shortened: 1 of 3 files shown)


cdeil:
@keflavich - Thanks!

Long-term I think an option to "ignore", "warn" or "error" on this would be best, like what the Python unicode encode / decode functions do.

For now: can you please add one test for one case where the warning is raised, asserting that it works?

keflavich:
maybe kwarg should be `skipped={'warn','ignore','error'}`?  What keyword do you prefer?

cdeil:
> maybe kwarg should be skipped={'warn','ignore','error'}? What keyword do you prefer?

I like `errors` for the name of the `parse` function argument, meaning "how to deal with parsing errors"

That's what Python uses for `str.encode` and `bytes.decode`:
```
In [6]: str.encode?
Docstring:
S.encode(encoding='utf-8', errors='strict') -> bytes

Encode S using the codec registered for encoding. Default encoding
is 'utf-8'. errors may be given to set a different error
handling scheme. Default is 'strict' meaning that encoding errors raise
a UnicodeEncodeError. Other possible values are 'ignore', 'replace' and
'xmlcharrefreplace' as well as any other name registered with
codecs.register_error that can handle UnicodeEncodeErrors.
Type:      method_descriptor
```
and
```
In [8]: bytes.decode?
Signature: bytes.decode(self, /, encoding='utf-8', errors='strict')
Docstring:
Decode the bytes using the codec registered for encoding.

encoding
  The encoding with which to decode the bytes.
errors
  The error handling scheme to use for the handling of decoding errors.
  The default is 'strict' meaning that decoding errors raise a
  UnicodeDecodeError. Other possible values are 'ignore' and 'replace'
  as well as any other name registered with codecs.register_error that
  can handle UnicodeDecodeErrors.
Type:      method_descriptor
```

regions/io/read_ds9.py
```diff
@@ -192,20 +194,33 @@ def ds9_region_list_to_objects(region_list):
             else:
                 raise ValueError("No central coordinate")
         else:
+            # Note: this should effectively never happen, because it would
```
cdeil: I think this will happen all the time. People have invalid `ds9` files and `region_type`. They mistype or whatever.

Can you change this code comment from "this should effectively never happen, ..." to something like "For invalid inputs, we do ..."

Is it possible to give the user a line number where the problem in their file occurs?
It will be really common and users will appreciate a helpful message.

regions/io/read_ds9.py
```diff
@@ -192,20 +194,33 @@ def ds9_region_list_to_objects(region_list):
...
+def ds9_string_to_objects(region_string, warn_skipped=False):
     """Parse ds9 region string to region objects
 
     Parameters
     ----------
     region_string : str
         DS9 region string
+    warn_skipped : bool
+        Print a warning if there is a skipped (commented) line?
+        Can set to ``False`` or ``'raise'`` if you want an exception instead.
```
cdeil: Like I said in the other comment, I'd prefer to call it "error".
And as-is, the option is documented as `bool`, but you also accept a string `raise`.
That's no good.
Let's make it a multi-choice string option.

regions/io/read_ds9.py
```diff
@@ -357,6 +385,19 @@ def line_parser(line, coordsys=None):
                 language_spec[region_type] = itertools.chain((coordinate, coordinate), itertools.cycle((radius,)))
 
             return region_type, parsed_return, parsed_meta, composite, include
+    else:
+        # it is not clear when it is OK to warn
```
cdeil: Can we resolve / improve the discussion on when to warn in this PR, and then change the comment?
It currently reads as an open issue "it is not clear ..."

Maybe just put whatever behaviour you like best, and establish it via tests?

regions/io/read_ds9.py
keflavich: I think this error actually *can't* occur by user error.  If a user mistypes / enters an incorrect region type, the previous function, `ds9_string_to_region_list`, should fail.  Only regions that are parsed by `ds9_string_to_region_list` but do not have a corresponding entry within `ds9_region_list_to_objects` should reach this point, and that should only happen if someone updating this code forgets to add a new entry in this function.

keflavich:
fyi, I'm pushing failing builds because I can't run tests locally; they all raise:
`E           OSError: Could not find lib c or load any of its variants [].`

cdeil:
> E OSError: Could not find lib c or load any of its variants [].

If you can't figure it out, maybe try `git clean -fdx` and `python setup.py test` in `master` and paste the full output in a gist, and I'll try to help by having a look at the log.

keflavich:
I can't figure it out, the problem persists, but I can run the tests directly now with
```
py.test regions/io/tests/test_ds9_language.py
```

keflavich:
@cdeil I *think* this is ready for final review now

cdeil:
@astrofrog or @joleroi - Do you have time for a quick review here?

cdeil:
I think it's good that you introduce a specific `DS9RegionParserWarning`.
But one motivation for custom warning / error is that people can catch it, right?
So it should be added to `__all__` and show up in the docs and be part of our public API, no?

For equivalent errors, should we stick with `ValueError` as you do here, or introduce `DS9RegionParserError`, to be equivalent with the warning?

regions/io/tests/test_ds9_language.py
```diff
@@ -1,17 +1,24 @@
...
+from astropy.utils.exceptions import AstropyUserWarning
+
 from ...shapes.circle import CircleSkyRegion
-from ..read_ds9 import read_ds9, ds9_string_to_objects
+from ..read_ds9 import read_ds9, ds9_string_to_objects, global_parser
 from ..write_ds9 import write_ds9, ds9_objects_to_string
 
-_ASTROPY_MINVERSION = v.LooseVersion('1.1')
-_ASTROPY_VERSION = v.LooseVersion(astrov.version)
+_ASTROPY_MINVERSION = vers.LooseVersion('1.1')
```
cdeil: We are working towards putting this as `astropy.regions` in the core package.
So there's no need to support old Astropy versions.

At the moment `regions` has to work with Astropy 1.2, and we can switch to 1.3 as soon as that is out.

So remove this `xfail` for Astropy < 1.1 here.

keflavich: I don't see any reason to remove the xfail now, as long as `regions` is still an independent package.  Sure, when we merge it into astropy-core, then it makes sense... but that's not now.

keflavich:
I like the idea of adding a `DS9RegionParserError` as a subclass of `ValueError`.... I'll do it if I have time.

cdeil:
I'm merging this now.

Thanks, @keflavich !

PR #92, opened 2016-12-14
Some improvements to BoundingBox

larrybradley:
These are mostly API doc changes.  I also added an example for `as_patch()`.

### DIFF
--- a/regions/core/bounding_box.py
+++ b/regions/core/bounding_box.py
@@ -81,7 +81,7 @@ class BoundingBox(object):
         ----------
         xmin, xmax, ymin, ymax : float
             Float coordinates defining a rectangle.  The lower values
-            (``min`` and ``min``) must not be greater than the
+            (``xmin`` and ``ymin``) must not be greater than the
             respective upper values (``xmax`` and ``ymax``).
 
         Returns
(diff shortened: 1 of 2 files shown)


regions/core/bounding_box.py
```diff
@@ -153,14 +164,31 @@ def extent(self):
 
     def as_patch(self, **kwargs):
         """
-        Return a Matplotlib patch that represents the bounding box.
+        Return a `matplotlib.patches.Patch` that represents the bounding
+        box.
```
cdeil: Link to the more explicit `matplotlib.patches.Rectangle` here instead of `matplotlib.patches.Patch`?
Put this info in a `Returns` section?

larrybradley: @cdeil Thanks!  Yes and yes.  Done.

regions/core/bounding_box.py
```diff
@@ -10,43 +10,49 @@
...
-    # Extent is useful when plotting bbox with matplotlib
+    # "extent" is useful when plotting the BoundingBox with matplotlib
     >>> bbox.extent  # matplotlib order: (x, y)
     (0.5, 9.5, 1.5, 19.5)
     >>> print(bbox.as_patch())
     Rectangle(0.5,1.5;9x18)
     """
 
     def __init__(self, ixmin, ixmax, iymin, iymax):
+        if ixmin > ixmax:
```
astrofrog: Should ``ixmin`` be strictly less than ``ixmax`` since the upper limit is exclusive? (that is, the smallest possible box is if ``ixmax == ixmin + 1``.

larrybradley: I think you are essentially asking if we should allow ``ixmin == ixmax``.  I left that possibility, but of course it is a `BoundingBox` of size 0 (along such axis).  If we don't want that, then I need to change this line to ``if ixmin >= ixmax`` (and the corresponding test for `y`).

cdeil: I don't see any pros / cons at the moment to allowing the size 0 case of `ixmin == ixmax`.
So I'm merging this now.

@astrofrog and all - feel free to make follow-up PRs proposing changes to this check any time.

cdeil:
Thanks, @larrybradley !

PR #93, opened 2016-12-15
Add clarification to BoundingBox extent docs

larrybradley:
This is a gentle suggestion to simply rename the `extent` property to `edges`.  When I see `extent` I immediately think it's a "size" of some kind, e.g. 3 meters in extent.  I think `edges` is a better name, since it's returning the edges of the bounding box and not a "size".  @astrofrog?

### DIFF
--- a/regions/core/bounding_box.py
+++ b/regions/core/bounding_box.py
@@ -148,11 +148,15 @@ class BoundingBox(object):
     @property
     def extent(self):
         """
-        The extent of the mask, defined as the bounding box from the
-        bottom-left corner of the lower-left pixel to the upper-right
-        corner of the upper-right pixel.
+        The extent of the mask, defined as the ``(xmin, xmax, ymin,
+        ymax)`` bounding box from the bottom-left corner of the
+        lower-left pixel to the upper-right corner of the upper-right
+        pixel.
 
-        This can be used for example when plotting using Matplotlib.
+        The upper edges here are the actual pixel positions of the
+        edges, i.e. they are not "exclusive" indices used for python
+        indexing.  This is useful for plotting the bounding box using
+        Matplotlib.
         """
 
         return (


cdeil:
I'm not even sure it's a useful property and would even be OK to remove it.

Here's what we have now:
http://astropy-regions.readthedocs.io/en/latest/api/regions.BoundingBox.html

The reason we added it was to be able to plot the bbox. Now there's as_patch.

Long term I could imagine we add something like [footprint](http://docs.gammapy.org/en/latest/api/gammapy.image.SkyImage.html#gammapy.image.SkyImage.footprint) or `BoundingBox.to_poly` that returns a little more useful representation that you can easily convert to sky coords as well.

Short term I have no opinion on what to put.
Maybe let's just do the rename suggested here.
@astrofrog, @larrybradley - Thoughts?

astrofrog:
Ah but I thought it was also used for the ``extent`` option in matplotlib, to center the image on the correct coordinates (not just to plot the box).

larrybradley:
The only place I see it being used is in the `as_patch()` method (and I just added that in #92), defining the lower-left edge of the patch.

astrofrog:
Sorry, just to clarify, even though it's not used *currently* in the documentation, it's been very useful for me when calling ``imshow`` with the ``extent`` keyword. I don't object to the renaming though, but I'd object to the removal.

larrybradley:
Ah, OK.  I was unaware that `imshow` has an `extent` keyword (I've never used that).  Given that this the name in `matplotlib`, I don't think we should rename this (even though I don't like the name :smile:).

larrybradley:
However, I will modify the PR to include the bit of clarification in the docs that `extent` values are the edges and not exclusive slice indices.

astrofrog:
By the way, I do actually use ``extent`` from ``imshow`` in http://astropy-regions.readthedocs.io/en/latest/getting_started.html#masks

larrybradley:
I modified this PR to leave `extent` alone, but added some clarification to the docs.

PR #97, opened 2017-01-06
Bounding box edges must be scalar

keflavich:
Prior to this fix, attempting to make a cutout resulted in errors like this:

```
File "<ipython-input-282-00fe094081c8>", line 7, in <module>
cutout = mask.cutout(data)
File "/Users/adam/repos/astropy-regions/regions/core/mask.py", line 154, in cutout
    cutout = data[self.bbox.slices]
TypeError: only integer scalar arrays can be converted to a scalar index
```

### DIFF
--- a/regions/core/bounding_box.py
+++ b/regions/core/bounding_box.py
@@ -105,10 +105,10 @@ class BoundingBox(object):
         BoundingBox(ixmin=1, ixmax=11, iymin=2, iymax=12)
         """
 
-        ixmin = np.floor(xmin + 0.5).astype(int)
-        ixmax = np.floor(xmax + 1.5).astype(int)
-        iymin = np.floor(ymin + 0.5).astype(int)
-        iymax = np.floor(ymax + 1.5).astype(int)
+        ixmin = int(np.floor(xmin + 0.5))
+        ixmax = int(np.floor(xmax + 1.5))
+        iymin = int(np.floor(ymin + 0.5))
+        iymax = int(np.floor(ymax + 1.5))
 
         return cls(ixmin, ixmax, iymin, iymax)
 


larrybradley:
@keflavich This change is fine, but I'm a bit puzzled as to how this was triggered.  The `_from_float` inputs should all be (regular, non-numpy) scalars.  Did you have scalar `ndarray` inputs (e.g. `xmin=np.array([10.5])`)?  Can you give an example that failed?  Thanks.

keflavich:
In short:
```
        phot_reg = regions.CircleSkyRegion(center=reg.center, radius=0.5*u.arcsec)
        pixreg = phot_reg.to_pixel(mywcs)
      
        mask = pixreg.to_mask()
        cutout = mask.cutout(data) * mask.data

```
where `reg` is a single region:
```
In [49]: reg
Out[49]:
PointSkyRegion
center: <FK5 Coordinate (equinox=J2000.000): (ra, dec) in deg
    ( 266.83354583, -28.38691944)>
```

So something is converting the scalar values of the coordinate and/or the radius to an array.  I didn't find exactly where that was.

larrybradley:
Thanks, I'll take a look next week.

larrybradley:
@keflavich  I can't reproduce this.  In your code, what is the `reg.center` input and what do you get for `mask.bbox`?

As a test, what happens when you do this:

```
>>> from regions import make_example_dataset
>>> dataset = make_example_dataset(data='simulated')
>>> wcs = dataset.wcs

>>> import astropy.units as u
>>> from astropy.coordinates import SkyCoord
>>> from regions import CircleSkyRegion

>>> center = SkyCoord(42, 43, unit='deg')
>>> radius = 5. * u.deg
>>> region = CircleSkyRegion(center, radius)
>>> pixreg = region.to_pixel(wcs)
>>> mask = pixreg.to_mask()
>>> mask.bbox
>>> BoundingBox(ixmin=42, ixmax=54, iymin=65, iymax=77)
>>> data = np.random.random((100, 100))
>>> cutout = mask.cutout(data) * mask.data
>>> cutout.shape
(12, 12)
```

larrybradley:
Also, in case you aren't aware the mask has a `multiply()` method that does`cutout = mask.cutout(data) * mask.data`:

```
>>> weighted_cutout = mask.multiply(data)
```

keflavich:
The problem is that the `Coordinate` object has single-valued array entries instead of scalar entries, e.g.:

```
>>> from regions import make_example_dataset
>>> dataset = make_example_dataset(data='simulated')
>>> wcs = dataset.wcs

>>> import astropy.units as u
>>> from astropy.coordinates import SkyCoord
>>> from regions import CircleSkyRegion,PointSkyRegion

>>> center = SkyCoord([42], [43], unit='deg')
>>> radius = 5. * u.deg
>>> pointreg = PointSkyRegion(center)
>>> region = CircleSkyRegion(pointreg.center, radius)
>>> pixreg = region.to_pixel(wcs)
>>> mask = pixreg.to_mask()
>>> mask.bbox
#BoundingBox(ixmin=42, ixmax=54, iymin=65, iymax=77)
>>> data = np.random.random((100, 100))
>>> cutout = mask.cutout(data) * mask.data  # FAILS HERE
>>> cutout.shape
#(12, 12)
```

I think this is driven by an error in the ds9 region reader, which I'll now look into

keflavich:
Also, yes, I'm aware of the `multiply` method, but I find `* mask.data` just as readable; is there any reason to prefer one over the other?

larrybradley:
OK, I get the same failure with `center = SkyCoord([42], [43], unit='deg')`.  In that case, I get
`BoundingBox(ixmin=[42], ixmax=[54], iymin=[65], iymax=[77])` -- "scalar" arrays as I suspected.

I think we should still merge this PR, but perhaps also add some validation checking for the region centers and the `BoundingBox` class to ensure the inputs are not arrays/lists.

larrybradley:
`multiply()` does exactly the same thing as `mask.cutout(data) * mask.data`, so it's your preference.

larrybradley:
Now that we figured this out, I'm going to merge this.

PR #98, opened 2017-01-09
ds9 reader: scalar coordinates

keflavich:
This is a minor fix to ds9 regions that forces regions with a single coordinate
(circle, point, etc.) to have scalar coordinates.

This will *also* fix the issue noted in #97


cdeil:
@keflavich - Is this ready to merge?

(I'd like to make the regions 0.2 release now)

PR #99, opened 2017-01-09
Make BoundingBox inputs integer

larrybradley:
This is a followup to #97 to ensure the inputs to `BoundingBox` are integer.

Note that `int()` is called on the inputs, which means floats will be silently truncated.  Do we also want to raise an exception for floats where e.g., ``int(ixmin) != ixmin``?

### DIFF
--- a/regions/core/bounding_box.py
+++ b/regions/core/bounding_box.py
@@ -3,6 +3,7 @@ from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 
 import numpy as np
+from astropy.io.fits.util import _is_int
 
 
 __all__ = ['BoundingBox']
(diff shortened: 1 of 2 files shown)


keflavich:
For consistency with numpy's new behavior, yes, we should raise an exception there.  I preferred the old approach of just warning, which I used all the time, but it will probably benefit many people... including me... to be forced to select how floats are truncated.

cdeil:
+1 to be strict on inputs and raise errors. Numpy now raises warnings where floats are used as indices (because I guess they can't throw errors to keep backward compat).

keflavich:
@cdeil I'm using numpy 1.12-dev, and I get exceptions if I try to use floats:

```
In [6]: np.__version__
Out[6]: '1.12.0.dev0+47c18a0'

In [7]: x = np.arange(5)

In [8]: x[2.0:3.0]
  File "<ipython-input-8-82b00ecaa918>", line 1, in <module>
    x[2.0:3.0]
TypeError: slice indices must be integers or None or have an __index__ method
```

larrybradley:
OK, done.  Input floats (even e.g. 3.0) will now raise a `TypeError`.

cdeil:
I'm still on Numpy 1.11 ... there it's a `VisibleDeprecationWarning`:
```
In [1]: import numpy as np

In [2]: np.__version__
Out[2]: '1.11.3'

In [3]: x = np.arange(5)

In [4]: x[2.0:3.0]
/opt/local/bin/ipython:1: VisibleDeprecationWarning: using a non-integer number instead of an integer will result in an error in the future
  #!/opt/local/Library/Frameworks/Python.framework/Versions/3.5/bin/python3.5
Out[4]: array([2])
```

PR #102, opened 2017-01-31
Fix ds9 metadata parsing: locals override globals

keflavich:
Previously, globals were incorrectly overriding locally specified metadata (color, linestyle, etc)


keflavich:
python2.7 thinks we're using unicode.  Not sure of the right workaround just yet...

cdeil:
travis-ci passed. Merge?

PR #106, opened 2017-02-14
Improve pixel region contains imlementation and tests

cdeil:
This pull requests contains some clean-up and tests for the pixel region contains implementation.
For now contains is only implemented for circle and polygon, and I'd like to leave the other regions to future PRs.

One thing I would like to fix here is to make polygon contains work with 2-dim PixCoord objects.
This is working for circles, and we're using this functionality in Gammapy here:
http://docs.gammapy.org/en/latest/_modules/gammapy/image/core.html#SkyImage.region_mask

@astrofrog @larrybradley - Do you have time to review this?

Any advice on how to make `PixelPolygonRegion.contains` work for 2-dim `pixcoord` inputs?
I could store the shape, flatten, call `points_in_polygon`, and reshape the output.
Or maybe `points_in_polygon` can be rewritten by just removing the `ndim=1` constraint, and make it work for any input arrays? (really it's a per-element computation, and the shape doesn't matter for the computation)

I'll start by adding unit tests for 2-dim PixCoord, and wait for your feedback on this point before changing `points_in_polygon`.

### DIFF
--- a/regions/shapes/tests/test_polygon.py
+++ b/regions/shapes/tests/test_polygon.py
@@ -2,12 +2,20 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 import numpy as np
 from numpy.testing import assert_allclose, assert_equal
-from astropy.tests.helper import pytest
+from astropy.tests.helper import pytest, assert_quantity_allclose
 from astropy import units as u
 from astropy.coordinates import SkyCoord
+from ..._utils.examples import make_example_dataset
 from ...core import PixCoord, BoundingBox
 from ..polygon import PolygonPixelRegion, PolygonSkyRegion
 from .utils import ASTROPY_LT_13, HAS_MATPLOTLIB
+
+
+@pytest.fixture(scope='session')
+def wcs():
+    config = dict(crpix=(18, 9), cdelt=(-10, 10), shape=(18, 36))
+    dataset = make_example_dataset(config=config)
+    return dataset.wcs
 
 
 class TestPolygonPixelRegion:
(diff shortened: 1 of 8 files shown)


cdeil:
So I did go a little further here and also implemented contains and pix <-> sky for polygons here and added a bunch of unit tests.

If someone has time to review this, or even contribute commits to this branch directly to make things better, please do!

regions/shapes/tests/test_polygon.py
```diff
@@ -1,52 +1,109 @@
...
+        # Bounding box and output shape is independent of subpixels,
+        # so we only assert on it once here, not in the other cases below
+        assert mask.bbox == BoundingBox(ixmin=0, ixmax=3, iymin=0, iymax=4)
+        assert mask.data.shape == (4, 3)
 
-        coords = PixCoord([13, 8], [15, 15])
-        assert_equal(poly1.contains(coords), [False, True])
+        # This example is with the default: subpixels=5
+        mask = self.reg.to_mask(mode='subpixels')
+        assert_allclose(np.sum(mask.data), 3.48)
```
cdeil: This polygon mask test is failing on travis-ci:
https://circleci.com/gh/astropy/regions/111?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

I kinda expected this, because pixels can land on the edge and results jump.
Probably I should just remove this check for `np.sum(mask.data)` and also all the ones below?
(those are still coverered by `pytest-array` tests checking masks I think, so should be OK, no?)

cdeil: Fixed in 9efee4f

PR #109, opened 2017-02-15
Rename Mask.__init__ argument from mask to data?

cdeil:
Currently the Mask API docs look like this:
http://astropy-regions.readthedocs.io/en/latest/api/regions.Mask.html

Looking at that I was confused, because the argument in `__init__` is called `mask`, and it's not visible that it's stored under the attribute name `data`.

@astrofrog @larrybradley - How about renaming the argument to `data`?

What's the best way to make the attribute `data` appear in the API docs?
I think  an `Attributes` section in the class docstring is needed, or we need to make it a property?

### DIFF
--- a/regions/core/mask.py
+++ b/regions/core/mask.py
@@ -10,7 +10,7 @@ class Mask(object):
 
     Parameters
     ----------
-    mask : array_like
+    data : array_like
         A 2D array of a region mask representing the fractional overlap
         of the region on the pixel grid.  This should be the full-sized
         (i.e. not truncated) array that is the direct output of one of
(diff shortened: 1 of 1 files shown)


cdeil:
I've attached a commit (00369c5) making the change.

@astrofrog @larrybradley - Yes? No?

larrybradley:
@cdeil  This change fine with me.

I think the easiest way to put `data` in the API docs is to simply add an Attributes section.  Making `data` a property just to get it in the docs doesn't make much sense to me.

cdeil:
Forgot to merge this one. Doing it now.

PR #114, opened 2017-02-23
Use exact minimal bounding box for rectangular region

larrybradley:
The current bounding box for rectangular regions is defined to contain the rectangle at all rotation angles.

The minimal bounding box here is defined for the "exact" rectangular region.  Note that for a mask defined with `mode="center"`, the bounding box could have an extra pixel of padding on each side.


larrybradley:
Example:

The red outline is the region.  The cyan outline is the bounding box.  The image data is the region mask.

Old:
old

New:
new

cdeil:
👍

PR #120, opened 2017-05-04
Add _repr_params to CompoundRegions

joleroi:
The ``__str__`` method of the compound sky and pixel region was broken, because it is implemented in the base class and calls into an attribute ``_repr_params`` which was not defined for the compound regions.

### DIFF
--- a/regions/core/compound.py
+++ b/regions/core/compound.py
@@ -14,6 +14,10 @@ class CompoundPixelRegion(PixelRegion):
         self.region1 = region1
         self.region2 = region2
         self.operator = operator
+        self._repr_params = [('component 1', self.region1),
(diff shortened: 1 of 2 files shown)


joleroi:
This seems to be uncontroversial :sweat_smile: - merging

cdeil:
@joleroi - I think it would have been nice to assert on a full `str` in the test.
IMO it's better because it
- would have let me see what the `str` / `repr` looks like by reviewing the diff for this PR on Github without having to create an example locally
- let's us notice when we accidentally change something that affects this str representation in the future.
- is just an extra line of typing

@joleroi - Could you please do this change to the test and link to the commit from this issue?

PR #121, opened 2017-05-04
Implement Rectangle sky->pixel transformation

keflavich:
Needs tests and sanity checks.

### DIFF
--- a/regions/shapes/rectangle.py
+++ b/regions/shapes/rectangle.py
@@ -3,9 +3,13 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 
 import numpy as np
 from astropy import units as u
+from astropy.coordinates import SkyCoord
 
-from ..core import PixelRegion, SkyRegion, Mask, BoundingBox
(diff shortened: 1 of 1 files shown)


cdeil:
I only had a quick look ... here's my comments:

 Yes, definitely needs a test. Ideally one comparing against what DS9 does, i.e. use it (possibly via the GUI) to do the pix <-> world transform for one example and check that this gives the same result (or that we understand and want the difference).

Also, it would be nice to add the other direction of the transform in the same PR and to add a test showing that it roundtrips, no?

The `to_polygon` is ~ 10 lines of formulae and it's not obvious what they do. Can you please add a comment at the top stating what they do (a rotation?)? Doesn't Astropy have utils functions somewhere to do such computations that would make the implementation more readable?

@astrofrog - Assigning you for review for now. If you don't have time, feel free to assign over to me or one of the other regions developers.

astrofrog:
Closing since this is no longer needed following #132

PR #126, opened 2017-05-08
Fix incorrect region number counting in regions_pyregi…

leejjoon:
in dev/regions_pyregion_comparison.py, the number of regions are calculated by counting the occurence of "(". This overestimates for the following cases

> panda(18:13:25.141,+31:21:31.32,51.8442,90.2623,1,0",4.995",1) # panda=(51.8442 90.2623 180.262 270.262)(0" 4.995" 9.99") text={Panda 2}

The patch only counts "(" not preceded by "=" or ")".

### DIFF
--- a/dev/regions_pyregion_comparison.py
+++ b/dev/regions_pyregion_comparison.py
@@ -18,9 +18,12 @@ from regions import read_ds9, write_ds9, DS9RegionParserError
 import pyregion
 import timeit
 import numpy as np
+import re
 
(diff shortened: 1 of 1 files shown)


keflavich:
Thanks for noticing this.  I was trying to find an example where overcounting would happen and I didn't see any in the test data we had, but this would also be a problem for any region with `(` in the `text={}` box.

leejjoon:
Yes, the proposed change can still count "(" in the text box. But as far as the coverage test is concerned, none of them has this.

Issue #45, opened 2016-06-26
Correctness and speed comparison to pyregion


cdeil:
As mentioned by @keflavich  in #9, it would be nice to as some point have a

- [ ] correctness comparion with pyregion
- [ ] speed comparison with pyregion

to make sure this `regions` package is a full replacement.

cdeil:
@joleroi started working on this in #123. Assigning this issue to you.

cdeil:
Related issue in the pyregion tracker: https://github.com/astropy/pyregion/issues/48

joleroi:
I will use this issue to come up with a TODO list for the regions parser. With [this tool](https://github.com/astropy/regions/blob/master/dev/regions_parse.py) you can easily try to parse a given ds9 example regions file. This is an incomplete list of issues I just discovered

* [ ] Implement region ``line``
* [ ] Implement region ``tile``

tbc

cdeil:
I'm moving this from v0.3 to the whishlist milestone, and unassigning @joleroi as he won't have time to work on this. If someone wants to make this a priority or has time to work on it, feel free to change the milestone or re-assign.

PR #129, opened 2017-05-09
Add Line{Circle|Sky}Region

joleroi:
One step towards fixing https://github.com/astropy/regions/issues/45

### DIFF
--- a/regions/shapes/line.py
+++ b/regions/shapes/line.py
@@ -1,0 +1,138 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+from astropy.wcs.utils import pixel_to_skycoord, skycoord_to_pixel
+
+from ..core import PixCoord, PixelRegion, SkyRegion, BoundingBox
+
+__all__ = ['LinePixelRegion', 'LineSkyRegion']
+
+
+class LinePixelRegion(PixelRegion):
+    """
+    A line in pixel coordinates.
+
+    Parameters
+    ----------
+    start : `~regions.PixCoord`
+        Start position
+    end : `~regions.PixCoord`
+        End position
+    """
+
+    def __init__(self, start, end, meta=None, visual=None):
+        self.start = PixCoord._validate(start, name='start', expected='scalar')
+        self.end = PixCoord._validate(end, name='end', expected='scalar')
+        self.meta = meta or {}
+        self.visual = visual or {}
+        self._repr_params = [('start', self.start),
+                             ('end', self.end)]
+
+    @property
+    def area(self):
+        """Region area (float)."""
+        return None
+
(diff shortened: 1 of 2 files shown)


regions/shapes/line.py
```diff
@@ -0,0 +1,138 @@
...
+    Parameters
+    ----------
+    start : `~regions.PixCoord`
+        Start position
+    end : `~regions.PixCoord`
+        End position
+    """
+
+    def __init__(self, start, end, meta=None, visual=None):
+        self.start = PixCoord._validate(start, name='start', expected='scalar')
```
keflavich: Should there be a better way to instantiate `PixCoord`s?  It seems weird to use private methods (`_validate`) here

regions/shapes/line.py
```diff
@@ -0,0 +1,138 @@
...
+    Parameters
+    ----------
+    start : `~astropy.coordinates.SkyCoord`
+        Start position
+    end : `~astropy.coordinates.SkyCoord`
+        End position
+    """
+    def __init__(self, start, end, meta=None, visual=None):
+        # TODO: test that start, end is a 0D SkyCoord
+        self.start = start
```
keflavich: Should we merge without having the `is_coord` check?  Or just add it?

regions/shapes/line.py
```diff
@@ -0,0 +1,138 @@
...
+        self.end = end
+        self.meta = meta or {}
+        self.visual = visual or {}
+        self._repr_params = [('start', self.start),
+                             ('end', self.end)]
+
+    @property
+    def area(self):
+        """Region area (`~astropy.units.Quantity`)"""
+        return None
```
keflavich: `0*u.sr`?

regions/shapes/line.py
```diff
@@ -0,0 +1,138 @@
...
+            start = PixCoord(start_x, start_y)
+            end_x, end_y = skycoord_to_pixel(self.end, wcs=wcs)
+            end = PixCoord(end_x, end_y)
+            return LinePixelRegion(start, end)
+
+        elif mode == 'affine' or mode == 'full':
+            raise NotImplementedError()
+
+        else:
+            raise ValueError('mode should be one of local/affine/full')
```
keflavich: is mode irrelevant here?

keflavich: it is not!  All modes do the same.

regions/shapes/line.py
joleroi: This (and also the PixCoord._validate issue) are unrelated to this PR - the same issue is present for all regions at the moment. So I would go for merging this and fixing the input validation separately.

joleroi:
@keflavich , others : Currently the line ``__init__`` takes ``start`` and ``end``. For the parser it would be easier if this was actually a len 2 list of SkyCoords (like a Polygone with 2 vertices). Opinions?

Issue #107, opened 2017-02-14
Simplify region base classes


cdeil:
In December 2016 @joleroi @astrofrog and I had a Google hangout where we discussed `astropy.regions`, the minutes are here:
https://docs.google.com/document/d/1u-WjsP4wLWx4g1-om7imstjngOmwpAJQ8EyVZxvhxV4/edit#heading=h.wf1hdss3eefb

I dropped the ball on implementing the changes discussed there, but I'd like to come back to that now. Actually, if this is non-controversial, I'd like to make the changes this week, before the regions 0.2 release, because it's not a lot of work, just removing a bunch of stuff instead of leaving dummy methods that return `NotImplementedError` at the moment.

Before going ahead, I'd like to ask @astrofrog and @joleroi again, and also @keflavich and @larrybradley for input!

---

Basically we said that `area` and `contains` should be removed from the base sky region class:
http://astropy-regions.readthedocs.io/en/latest/api/regions.SkyRegion.html

The reason is that `area` and `contains` aren't well-defined concepts for all sky regions.
We argued about this for quite a while, but in the end I think there was agreement that it's not well-defined and thus doesn't belong in the base class.
E.g. how exactly would you implement `area` and `contains` for
http://astropy-regions.readthedocs.io/en/latest/api/regions.RectangleSkyRegion.html
so that it's valid for very large rectangles (say width = 100 deg)?

So we'd like to propose basically that we adopt the DS9 model of having regions mainly defined as pixel regions, and having sky regions mostly as convenience, but not with separate definitions of `contains` methods. Instead `contains` would be implemented in the following generic way on the base class for sky regions, and only be well-defined if the user has a WCS:

```
class SkyRegion(object):
    def contains(self, sky_coord, wcs):
         Reg = self.to_pixel(wcs)
         Pix_coord = foo(sky_coord)
         Return reg.contains(pix_coord)
```

In the future, we would add the special cases (circle & polygon) back where contains is well-defined in sky coordinates, either as separate classes or on the existing classes.

This brings us to the other major change we'd like to propose, which is to get rid of the different modes in the pixel to sky transformation:
- http://astropy-regions.readthedocs.io/en/latest/api/regions.SkyRegion.html#regions.SkyRegion.to_pixel
- http://astropy-regions.readthedocs.io/en/latest/api/regions.PixelRegion.html#regions.PixelRegion.to_sky

This was never implemented, it was just an idea, there will be no functionality lost.

[...]

Issue #86, opened 2016-11-30
Improve ellipse regions


cdeil:
I'd like to propose a few changes to the ellipse regions, [EllipsePixelRegion](http://astropy-regions.readthedocs.io/en/latest/api/regions.EllipsePixelRegion.html) and [EllipseSkyRegion](http://astropy-regions.readthedocs.io/en/latest/api/regions.EllipseSkyRegion.html), that I don't have time for now to implement in #85.

- [ ] Rename parameter `major` -> `a` and `minor` -> `b`. These are the names that were previously chosen for [astropy.modeling.functional_models.Ellipse2D](http://docs.astropy.org/en/latest/api/astropy.modeling.functional_models.Ellipse2D.html) and [photutils.aperture.EllipticalAperture](http://photutils.readthedocs.io/en/latest/api/photutils.aperture.EllipticalAperture.html).
- [ ] The ellipse mask test in `regions/shapes/tests/test_masks.py` currently uses `EllipsePixelRegion(PixCoord(3.981987, 4.131378), major=2.2233, minor=3.3411, angle=32 * u.deg)`, i.e. minor > major, probably due to a misunderstanding in the argument order. I suggest we change this test to use `a > b` and update the reference results files, even if `a` and `b` are no longer called major and minor.
- [ ] The description of `a` and `b` should not claim they are the major and minor axis. That's a can of worms, because sometimes `a > b` and `b < a`, we don't have any control over that in the code. This should also be fixed in the [astropy.modeling.functional_models.Ellipse2D](http://docs.astropy.org/en/latest/api/astropy.modeling.functional_models.Ellipse2D.html) docstring, I think there's already an issue with no action discussing that. We can expose extra attributes `major` and `minor` if people want, or a method that returns a new shape where always `a>b` if people think that's useful.

Just overall, I think it's important that we finish the photutils aperture -> astropy regions code transition soon and systematically and someone spend a week to do that. We're wasting time looking at the same issues that have already been discussed, sorted out and are coded up and tested in `photutils`. (I'm not saying we have to do everything the same, this is a chance to fix inconsistencies and mistakes we made in photutils if any.)

Putting this on the 0.3 milestone. Help wanted!

astrofrog:
Could we use ``width`` and ``height`` instead of ``a`` and ``b`` to be consistent with Rectangle and also the terminology ds9 uses?

cdeil:
@astrofrog - Assigning to you for a decision.

[...]

PR #132, opened 2017-05-11
Refactor sky regions

astrofrog:
This changes the current sky regions to be simply regions that represent the same shapes as the pixel regions but with properties specified in world coordinates (e.g. a circle in an image with a center given by a skycoord) and for now this removes the sphericalskyregion functionality for circles (will add this back after). This is to address https://github.com/astropy/regions/issues/107

This also implements some of the missing area calculations, to_shapely, to_sky, etc and I've split up the docs into several sections and improved it in places.

You can preview the updated docs at: http://astropy-regions-astrofrog.readthedocs.io/en/latest/

I plan to then try and add the full spherical sky polygon and circle in a separate PR, but as-is this should now provide full compatibility/consistency with ds9.


astrofrog:
@joleroi - here's the file I was using to run the integration test: https://www.dropbox.com/sh/pfh42u7mexazuo8/AABn40AUsgPB785g58FHZhiQa?dl=0

astrofrog:
Just FYI for anyone looking at this, we still need to do some work to check that the ellipse and rectangle are correctly rotated (they appear to be 90 degrees off at the moment compared to ds9)

regions/core/core.py
```diff
@@ -283,47 +259,29 @@ def union(self, other):
         from .compound import CompoundSkyRegion
         return CompoundSkyRegion(self, other, operator.or_)
 
-    @abc.abstractmethod
-    def contains(self, skycoord):
+    def contains(self, skycoord, wcs):
```
joleroi: Can we add ``__contains__`` calling this method also for SkyRegions (it's there for pixel regions)

joleroi: Ah no you can't because you have to pass a ``wcs`` ...

regions/core/pixcoord.py
```diff
@@ -143,6 +143,7 @@ def from_sky(cls, skycoord, wcs, origin=_DEFAULT_WCS_ORIGIN, mode=_DEFAULT_WCS_M
         Calls :meth:`astropy.coordinates.SkyCoord.to_pixel`.
         See parameter description there.
         """
+        print(skycoord)
```
joleroi: remove?

regions/core/core.py
astrofrog: Yep!

regions/core/pixcoord.py
astrofrog: Oops, thanks!

astrofrog:
I think this is basically ready for review at this point. I need to find a better way to do the Point patch, but I think that can be fixed later. There are also some FIXMEs and TODOs but none are critical but more for future.

regions/core/core.py
```diff
@@ -283,47 +259,29 @@ def union(self, other):
...
+        Check whether a sky coordinate falls inside the region
 
         Parameters
         ----------
         skycoord : `~astropy.coordinates.SkyCoord`
             The position or positions to check
+        wcs : `~astropy.wcs.WCS` instance
+            The world coordinate system transformation to assume
         """
+        from .pixcoord import PixCoord
```
cdeil: Move import to the top of the file.

cdeil:
The docs restructuring looks good to me. The section that sticks out now even more as useless is "What next?". I guess it could be removed. Or rewritten, e.g. the obvious thing to do next is look at the API docs for details, so that could be mentioned.

regions/core/core.py
astrofrog: I can't, because of circular imports

astrofrog:
Note that this now implements the change of ellipse parameters to width/height as discussed in https://github.com/astropy/regions/issues/86

astrofrog:
@cdeil - I'll work on making sure we have proper test for all the rotations

regions/core/core.py
cdeil: Ah, OK, I see we have `from .core import _DEFAULT_WCS_ORIGIN, _DEFAULT_WCS_MODE` in `pixcoord.py`.  This is not nice that we have circular imports between `pixcoord.py` and `core.py`. I'm +1 to move those constants to `pixcoord.py` to have it be standalone (and because all pixel / world transforms should go through `PixCoord` to abstract away this choice we've made for `regions`.

astrofrog:
One of the easiest ways to make sure that everything is consistent is to make plots like this:

everything

I was wondering whether we might want to consider adding image tests to the test suite to test this kind of all-ecompasing plot? Or would people prefer that we really just compare numerical results? (no strong feelings, but just want to know before implementing anything)

keflavich:
I like this sort of visual test, but I would prefer that it be kept optional so that I can run the tests locally without getting popup windows and faster.

astrofrog:
Ok - the main difficulty with testing this without plots is how to test e.g. that the shapely region is correct. I don't really want to have to write out all the array values for the ellipse edges. So what about adding a few image tests (keeping it to a minimum of one plot per shape), exclude axis labels (since that's usually what changes between matplotlib versions), making it so when tests are run the backend is set to Agg (can be done) to avoid pop-up windows? The tests should still be speedy.

astrofrog:
Another possibility is to simply write out the exact ellipse vertex arrays to ASCII files (like we do for masks) but that produces less visual checks.

cdeil:
+1 to separate test for computations (e.g. contains) from plotting tests.

@astrofrog - you mention that it's difficult to write the test for the Shapely objects.
Why do we need large arrays / masks there?
Can't you assert on the data member parameters of the Shapely ellipse object, and maybe in addition call contains on ~ 3 well-chose points that are inside and outside and establish the ellipse orientation?

astrofrog:
I was being lazy... I will implement proper tests, as I have some ideas on doing that efficiently

regions/core/core.py
astrofrog: Let's do that in a separate PR? This one is getting out of control :)

astrofrog:
Ok, so if tests pass, I would suggest we merge this and we can always tweak things in future PRs. This one is getting a bit out of control. I've refactored the tests to move a lot of the common logic to common base classes, and I've made it so that we test our contains method against shapely for hundreds of points (which gives me more confidence). We could still add more tests but I would prefer to merge first and add these in separate PRs.

docs/ds9.rst
```diff
@@ -0,0 +1,82 @@
...
+    >>> write_ds9(regions, filename)
+    >>> regions = read_ds9(filename)
+    >>> regions
+    [CircleSkyRegion
+     center: <FK5 Coordinate (equinox=J2000.000): (ra, dec) in deg
+         (245.3477, 24.4291)>
+     radius: 3.0 deg]
+
+Often DS9 region files contain extra information about colors or other attributes.
+This information is lost when converting to `~regions.Region` objects.
```
keflavich: This should not be true; the data are (or at least were) preserved in `.meta` and `.visual`

astrofrog: I don't want to modify this as it might conflict with changes in #130, so I think we should change it there once this PR is merged.

astrofrog:
@cdeil - any objections to merging this now? I plan to work on further improvements, but this PR is getting a little out of hand and already implements the original refactoring plan  (we should get #130 merged soon too)

cdeil:
Sorry, I didn't know what I was supposed to merge. Merging now.
(we can do further review in master and everyone that has time can do follow-up issues or PRs)

cdeil:
@astrofrog - Thank you!!!!

docs/ds9.rst
joleroi: Fixed in #130

keflavich:
@astrofrog This is a too-late comment, but I have a problem with the new API.  I have a set of regions that have been created as sky regions with no associated WCS: they are positions on the sky with circles around them representing regions of constant angular distance.  When I do `reg.contains(coordinate)`, it make no sense at all to require a WCS.

While I agree with the original premise, since many regions are undefinable without an associated projection (e.g., rectangles), I expect other users to encounter this case commonly as well.  The workaround now is hideous: I have to create an arbitrary WCS from scratch.  Do you have any alternative suggestions, or ideas how we can make this more sensible?

cdeil:
We discussed this quite a bit, and my understanding is this:
* There's very few sky regions where the shape (e.g. "contains") is well-defined without a WCS. I'm only aware of circle and polygon.
* The plan is to bring that "true sky region" functionality back as separate classes.
* Of course trying to add the "true sky region" case back into the existing circle and polygon sky region classes can be discussed again, but I think it's messy. In any case, there's some things that aren't so nice, e.g. the "true sky regions" aren't available in the DS9 serialisation format, i.e. DS9 just doesn't support them.

keflavich:
Yeah, it might only make sense to support those two regions, but we should.

In the circle case, the true sky region and WCS-based region approach should be equivalent, right?  Is it possible that the true sky region approach is also more computationally efficient?

astrofrog:
I'm going to add back the 'spherical' circle and polygon in a future PR.

> In the circle case, the true sky region and WCS-based region approach should be equivalent, right? Is it possible that the true sky region approach is also more computationally efficient?

No, a true spherical circle is not necessarily equivalent to the current approach, especially for images that span a large area of sky. But checking for points inside the true spherical circle might indeed be more efficient, though that remains to be tested.

joleroi:
> In the circle case, the true sky region and WCS-based region approach should be equivalent, right?

For completeness: I guess this was discussed  in https://github.com/astropy/regions/issues/76

bsipocz:
This PR splitting up the docs has a side effect of now the examples are failing, and the references in the narrative doesn't make sense (https://astropy-regions.readthedocs.io/en/latest/contains.html)

cdeil:
ref #263

PR #137, opened 2017-05-26
Implement to_pix and to_mask for compound regions

joleroi:
This PR implements
* ``to_pixel`` for ``CompoundSkyRegion``
* ``to_mask`` for ``CompoundPixelRegion``
* ``to_pixel`` for ``CircleAnnulusSkyRegion``, since I didn't know if it's possible to return a ``CircleAnnulusPixelRegion`` instance when calling the ``to_pixel`` method on the mother class (``CompoundSkyRegion``)\

### DIFF
--- a/regions/core/compound.py
+++ b/regions/core/compound.py
@@ -11,13 +11,17 @@ class CompoundPixelRegion(PixelRegion):
     Represents the logical combination of two regions in pixel coordinates.
     """
 
-    def __init__(self, region1, operator, region2):
+    def __init__(self, region1, region2, operator):
+        if not isinstance(region1, PixelRegion):
+            raise TypeError("region1 must be a PixelRegion")
+        if not isinstance(region2, PixelRegion):
+            raise TypeError("region2 must be a PixelRegion")
+        if not callable(operator):
+            raise TypeError("operator must be callable")
+
         self.region1 = region1
         self.region2 = region2
         self.operator = operator
-        if not callable(operator):
-            raise TypeError("The operator passed to a compound region must "
-                            "be callable.")
         self._repr_params = [('component 1', self.region1),
                              ('component 2', self.region2),
                              ('operator', self.operator),
@@ -38,7 +42,13 @@ class CompoundPixelRegion(PixelRegion):
             ixmin=min(mask1.bbox.ixmin, mask2.bbox.ixmin),
             ixmax=max(mask1.bbox.ixmax, mask2.bbox.ixmax),
             iymin=min(mask1.bbox.iymin, mask2.bbox.iymin),
-            iymax=max(mask1.bbox.iymax, mask2.bbox.iymax))
+            iymax=max(mask1.bbox.iymax, mask2.bbox.iymax)
+        )
+
+        bbox_borders = np.array([bbox.ixmin, bbox.ixmax, bbox.iymin, bbox.iymax]) 
+        if (bbox_borders < 0).any():
+            raise NotImplementedError("Bounding box must be within array for "
+                                      "compound regions, see ")
 
         # Pad mask1.data and mask2.data to get the same shape
         padded_data = list()
@@ -51,7 +61,7 @@ class CompoundPixelRegion(PixelRegion):
                                       ((ptop, pbottom), (pleft, pright)),
                                       'constant'))
 
-        data = self.operator(*np.array(padded_data, dtype=np.bool))
+        data = self.operator(*np.array(padded_data, dtype=np.int))
         return Mask(data=data, bbox=bbox)
 
     def to_sky(self, wcs):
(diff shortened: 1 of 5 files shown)


joleroi:
This will obviously only work for ``mode=center``, since I do boolean operations on the masks. I don't think the other cases (i.e. float values in the mask) are well defined for a compound region. Ok, to raise an error if ``mode`` is not ``center`` for compound regions, or does someone have an idea how to implement those?

astrofrog:
@joleroi - for other modes, I think the only solution is to make the regions into polygons and then use the exact mode there. But exact mode doesn't exist for polygons yet. Just raise NotImplementedError for now and we can decide how to deal with this later.

regions/core/compound.py
```diff
@@ -22,8 +23,30 @@ def __init__(self, region1, operator, region2):
...
+    def to_mask(self, mode='center', subpixels=1):
+        mask1 = self.region1.to_mask(mode=mode, subpixels=subpixels)
+        mask2 = self.region2.to_mask(mode=mode, subpixels=subpixels)
+
+        # Common bounding box
+        bbox = BoundingBox(
+            ixmin=min(mask1.bbox.ixmin, mask2.bbox.ixmin),
+            ixmax=max(mask1.bbox.ixmax, mask2.bbox.ixmax),
+            iymin=min(mask1.bbox.iymin, mask2.bbox.iymin),
+            iymax=max(mask1.bbox.iymax, mask2.bbox.iymax))
```
keflavich: Just a sanity check - are the bbox entries guaranteed to be positive?  They ought to be; if they are not, we'll get indexing like `[-1:10]` which result in empty.

cdeil:
@joleroi - Any chance you can finish this PR up (add minimal tests that execute one simple example case) in the coming days or weeks?

I think it would be really nice if we try to merge all the open PRs in and do another astropy-regions release soon (but no schedule yet).

cdeil:
@joleroi - Do you have time to finish this up or should I / someone take over?

regions/core/compound.py
joleroi: Shouldn't such a check be added [here](https://github.com/astropy/regions/blob/master/regions/core/bounding_box.py#L56)?

cdeil: @joleroi - I forgot how regions works - is `BoundingBox` always supposed to be within the array, or can it extend beyond and some clipping is supposed to occur later when BBoxes are applied?

Could you please check the existing tests for `to_mask` and see if / how it works if a region extends beyond the image? It should then behave the same for compound regions `to_mask`, i.e. succeed or fail in the same way if possible.

Let me know if it's not clear and we'll have a look together. I would start by adding a test case for that and just see what happens.

cdeil:
Rebase please (so that tests run on CI and it's easier to check it out locally).

regions/core/compound.py
joleroi: @cdeil Apparently, BoundingBoxes can extend over the array

```
In [1]: from regions import PixCoord, CirclePixelRegion

In [2]: reg = CirclePixelRegion(PixCoord(1,1), 5)

In [3]: reg.to_mask().bbox
Out[3]: BoundingBox(ixmin=-4, ixmax=7, iymin=-4, iymax=7)
```

I will try to make this behaviour reproducible for ``CompoundPixelRegions``

cdeil: Are there cases which are  obviously a bug or at least bad behaviour, where a user extracts a mask and a negative index is silently used for slicing, i.e. an orbitrary cutout is returned, but really an error should be raised or clipping to `i_min = ` should occur?

then you might have a bigger task on your hand here ...

cdeil: If you want to finish this up and put it in, we can, and you open a reminder issue on that question for future discussion / work.

joleroi:
Ready for review

regions/core/compound.py
```diff
@@ -25,10 +30,47 @@ def __init__(self, region1, operator, region2):
...
+            ixmax=max(mask1.bbox.ixmax, mask2.bbox.ixmax),
+            iymin=min(mask1.bbox.iymin, mask2.bbox.iymin),
+            iymax=max(mask1.bbox.iymax, mask2.bbox.iymax)
+        )
+
+        # Bounding boxes must not extend over array, see #168
+        bbox_borders = np.array([bbox.ixmin, bbox.ixmax, bbox.iymin, bbox.iymax]) 
+        if (bbox_borders < 0).any():
+            raise NotImplementedError("Bounding box must be within array for "
+                                      "compound regions, see ")
```
cdeil: Add ref to issue number here?

cdeil:
@joleroi - Thanks!

If someone else still has comments on this, I'd suggest to continue in #168 or in a new follow-up issue or pull request.

PR #143, opened 2017-06-23
ds9 reader: ellipses are stored with radii but read as width/height

keflavich:
The ds9 reader is getting ellipses wrong right now; it is interpreting the stored 'radius' as a diameter.

I'll try to figure out how to handle this.

### DIFF
--- a/regions/io/ds9/core.py
+++ b/regions/io/ds9/core.py
@@ -68,7 +68,7 @@ class Shape(object):
     """
     shape_to_sky_region = dict(
         circle=shapes.CircleSkyRegion,
-        ellipse=shapes.EllipseSkyRegion,
+        ellipse=shapes.Ds9EllipseSkyRegion,
         box=shapes.RectangleSkyRegion,
         polygon=shapes.PolygonSkyRegion,
         annulus=shapes.CircleAnnulusSkyRegion,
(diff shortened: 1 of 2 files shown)


astrofrog:
I am against creating a special Ds9 region class. We should settle on a single definition and just translate as needed during the Ds9 region file I/O.

keflavich:
I looked into modifying the reader, and it requires more special-case code to handle this at read time because of the way ellipses are stored in ds9 files.  An ellipse can be either a single ellipse or a list of consecutive ellipses, i.e., elliptical annuli, and there is no way to know which except by counting the number of inputs.

The approach I had attempted, but decided was inviable, was to add a `diameter` option in addition to the `radius` in the parser.  I'll revisit this, though, because it is probably technically possible.

cdeil:
@astrofrog wrote:

> I am against creating a special Ds9 region class. We should settle on a single definition and just translate as needed during the Ds9 region file I/O.

@keflavich wrote:

> An ellipse can be either a single ellipse or a list of consecutive ellipses, i.e., elliptical annuli, and there is no way to know which except by counting the number of inputs.

So if DS9 has crazy, complex regions, I don't think we should compromise on a simple and clean data model in `astropy.regions` just to support DS9. Obviously this is a difficult decision, we do want to support existing regions and workflows as much as possible.

So we now have a Shape and ShapeList that is like pyregion did it and can store all DS9 regions? Then maybe the DS9 interface can aim to fully support all regions at that level, i.e. the user can get Shape and ShapeList. But only a subset of regions can translate further into the regions objects like our ellipse regions?

PR #155, opened 2018-02-01
Update CircleCI config to use the manylinux 32-bit image

larrybradley:
With this configuration we don't need to support a 32-bit docker image for affiliate packages.


astrofrog:
@larrybradley - interesting that the CircleCI build is working here - is the issue you mentioned for photutils not present here?

larrybradley:
It was a `photutils` doctest that failed.  There isn't such a test in `regions`, but I can add one.

astrofrog:
@larrybradley - ok, it would be great if you could add a test like that here (I'd incorrectly assumed it was one of the geometry tests I copied over here that failed)

larrybradley:
@astrofrog I've added the test.  It fails for `CircleCI` (32-bit), but travis passes.

The AppVeyor failure is unrelated.  It looks like that is related to the numpy 1.14.0 bug with np.einsum.

astrofrog:
@larrybradley - I don't think this is really a bug - the issue is that there are some subpixel points that fall exactly on the edge of the circle because of the round numbers used for the center, radius, and subpixel ratio (specifically x=2.4 y=1.8, x=-2.4 y=1.8, and so on). So I would suggest removing the test here and instead adjusting the photometry test in photutils so that it isn't as sensitive to this issue.

larrybradley:
@astrofrog I removed the test here.  The AppVeyor failure is unrelated.

Issue #65, opened 2016-08-03
ds9 region parsing should handle whitespace


keflavich:
As per https://github.com/astropy/pyregion/issues/73, there are some regions that are valid for ds9 but are not included in the test cases now and are not properly parsed.

sushobhana:
hi @keflavich ,
i went to said issue and tried to reproduce the error. the problem i saw  that it doesn't identify `J2000` as a coordinate system. i think it should be mapped to `fk5` and it seems to handles whitespace pretty well ( i only tried to reproduce the example given: `'J2000; circle 188.5557102 12.0314056 1" # color=red' `). it seems to work after that..

keflavich:
If you think you've found a solution, yes, go ahead and open a PR.  I'm not entirely sure the J2000 frame is the problem, given the description in the other issue, but it might be.

sushobhana:
@keflavich initially i thought the same. But i built it and it's working.! Do have a look over the PR.

larrybradley:
This issue is fixed.

PR #157, opened 2018-02-22
mapped_J2000

sushobhana:
ISSUE: #65

### DIFF
--- a/regions/io/ds9/read.py
+++ b/regions/io/ds9/read.py
@@ -136,12 +136,13 @@ class DS9Parser(object):
       ``warn`` will raise a warning, and ``ignore`` will do nothing
       (i.e., be silent).
     """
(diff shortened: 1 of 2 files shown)


sushobhana:
output:
screenshot from 2018-02-22 23-04-53

keflavich:
@sushobhana Could you post text snippets instead of screenshots of text?

That said, this looks fine.  It would be helpful to add a regression test too.  Basically, add the example you screenshotted as a test

sushobhana:
will add the test soon.

sushobhana:
@keflavich is it ok to go?

keflavich:
No, there are several test failures.  However, I don't think they're related to what you did, so I've restarted the tests.  Let's see if they pass now.  If not, we may need to ask someone else for help.

sushobhana:
few blank spaces in the last line were causing the trouble..seems good now!

PR #165, opened 2018-03-21
shapely methods for regions.

sushobhana:
1. Added `to_shapely()` for CircleAnnulusPixelRegion.
2. Added `from_shapely()` for circle,point
3. Added tests for above.

### DIFF
--- a/regions/shapes/annulus.py
+++ b/regions/shapes/annulus.py
@@ -53,6 +53,11 @@ class CircleAnnulusPixelRegion(CompoundPixelRegion):
     def bounding_box(self):
         return self.region2.bounding_box()
 
+    def to_shapely(self):
+        r1 = self.region1.center.to_shapely().buffer(self.region1.radius)
+        r2 = self.region2.center.to_shapely().buffer(self.region2.radius)
+        return r2.symmetric_difference(r1)
+
 
 class CircleAnnulusSkyRegion(CompoundSkyRegion):
     """
--- a/regions/shapes/annulus.py
+++ b/regions/shapes/annulus.py
@@ -54,8 +54,8 @@ class CircleAnnulusPixelRegion(CompoundPixelRegion):
         return self.region2.bounding_box()
 
     def to_shapely(self):
-        r1 = self.region1.center.to_shapely().buffer(self.region1.radius)
-        r2 = self.region2.center.to_shapely().buffer(self.region2.radius)
+        r1 = self.region1.to_shapely()
+        r2 = self.region2.to_shapely()
         return r2.symmetric_difference(r1)
 
 
--- a/regions/shapes/annulus.py
+++ b/regions/shapes/annulus.py
@@ -37,7 +37,6 @@ class CircleAnnulusPixelRegion(CompoundPixelRegion):
                              ('outer radius', region2.radius),
                              ('center', region2.center)]
 
-
     @property
     def center(self):
         return self.region1.center
--- a/regions/shapes/circle.py
+++ b/regions/shapes/circle.py
@@ -44,6 +44,16 @@ class CirclePixelRegion(PixelRegion):
     def to_shapely(self):
         return self.center.to_shapely().buffer(self.radius)
 
+    @staticmethod
+    def from_shapely(circle):
+        from shapely.geometry import Polygon
+        if not isinstance(circle, Polygon):
+            raise TypeError('the given argument is not of type "shapely.geometry.polygon.Polygon".')
+        else:
+            center = PixCoord(circle.centroid.x, circle.centroid.y)
+            radius = circle.exterior.coords[0][0] - center.x
+            return CirclePixelRegion(center, radius)
+
     def to_sky(self, wcs):
         # TODO: write a pixel_to_skycoord_scale_angle
         center = pixel_to_skycoord(self.center.x, self.center.y, wcs)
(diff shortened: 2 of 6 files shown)


sushobhana:
is there any way we can skip the tests for CIs since shapely is not added as a dependency ? Or should i just remove them??

cdeil:
@astrofrog @keflavich @larrybradley - Are you convinced region shapely converters should be added?

I wasn't sure, given that the long-term plan here is `astropy.regions` in the core package and we probably want a small and core feature set / API, and the fraction of astronomers that need / want to work with Shapely is probably very small.

If you all think it should be added, OK. I just wanted to bring up the question and have it decided given that now @sushobhana is spending time to work on this.

keflavich:
@cdeil - fair question.  Allowing the (optional) link to other packages is relatively harmless.  However, I have a vague recollection that we *needed* shapely for certain operations that would be very expensive (in person-time) to reimplement.  I don't recall what these were, but I thought we had decided that using shapely for some operations was OK.

cdeil:
> Allowing the (optional) link to other packages is relatively harmless. 

Agreed. (but there is some cost to add it and then to support it for the next 10+ years)

> I thought we had decided that using shapely for some operations was OK.

Well, the main thing is cartesian polygon contains. But I think that was already implemented by @astrofrog or in any case, I wouldn't want to have a Shapely dependency for such a core operation.

Shapely has other things like union / intersect / polygonisation of shapes, padding of polygons ... fancy stuff, but something that only ~ 1% ? of astronomers need (I don't know anyone). And those experts presumably could write a converter to Shapely or whatever fancy geometry packages they need.

I realise my comments don't help much to make the decision about shapely, it's still a close call IMO whether we want to support it. I just wanted to mention that the thinking should be concerning long-term vision of `astropy.regions` - do we want to support shapely converters there?

keflavich:
At the very least, your comment raises the point that we should not *prioritize* the shapely converters.  @sushobhana, thanks for starting this contribution (it is a good demonstration for your GSoC application), but you should not spend too much time on this until/unless we decide that it is an important core component of the package.

sushobhana:
Well, i was just going through it's documentation and wrote it, that's why i opened a PR. I will not spend more time on it though.! It seems like it is not the need of the hour!

sushobhana:
> Shapely has other things like union / intersect / polygonisation of shapes, padding of polygons ... fancy stuff, but something that only ~ 1% ? of astronomers need (I don't know anyone). And those experts presumably could write a converter to Shapely or whatever fancy geometry packages they need.

I think to plot compound regions, the intersection/union of polygons are important. Since, `regions` is not completely independent of `shapely` , do you think we should add plotting methods for `Compund` regions using `shapely` , or we just leave it until we find some other alternative??
@keflavich

keflavich:
For the moment, the easy way out for compound regions is just to plot each region included in the compound; for now, I'd suggest going that route.

keflavich:
You can compound or composite regions in ds9 (using 'compound' and 'composite' as verbs there...); it displays as I described, where each region in the compound region is displayed.  So I think that's the right thing to do.  Similarly, as long as we're only dealing with boolean masks, just adding the masks of the two subsets should be fine.

```
# composite(+0:16:51.371,-0:00:56.025,58.688) || composite=1
circle(+0:17:19.224,-0:00:51.685,35.279") ||
circle(+0:16:23.518,-0:01:00.366,33.244")
```

cdeil:
So what does this mean for this PR? Close, or rebase and merge?

I'd be +1 to remove the shapely stuff completely, but it could go either way. So if you want to add this for now, OK, go ahead. (But I might bring the question / suggestion again later whether we should have the shapely methods, once `astropy.regions` is at the point where we consider proposing it for Astropy core.

sushobhana:
> For the moment, the easy way out for compound regions is just to plot each region included in the compound; for now, I'd suggest going that route.

You mean add individual patches to the axes? If yes, this would be ugly coz regions might overlap/invert.
Users can choose this approach if they want it manually.And this would not require `shapely` , if we would want to implement this for now.
 But in general, if we want to plot/work on compound regions, I think we need to keep `shapely` until  the polygon regions supports simple addition/subtraction completely.
So, to close the PR or work on it for now?

keflavich:
@cdeil @astrofrog Could we wrap up discussion on this PR while we still have @sushobhana's undivided attention?  We left the decision on whether we want this undecided.  The issues are:
(1) Shapely provides some unique functionality, but it's not clear there's any demand for this functionality.
(2) Adding `shapely` as a dependency creates a long-term maintenance burden and may raise questions during the inevitable (and perhaps imminent) merge into `astropy.regions`.

I'm still undecided, but I lean toward closing this for now but remaining aware that, if users need spatial overlap features, we would revive this PR to provide those.

cdeil:
I'm +1 to remove `to_shapely` and `from_shapely`, and to replace it with a docs page showing one or two examples how one can do these transforms and use shapely for some feature regions doesn't offer.

My arguments are:
- `astropy-regions` should become `astropy.regions` in the core package soon.
- from the user perspective: very few astronomers need or have ever heard of Shapely, support for it doesn't belong in Astropy core IMO, it's a feature for 1% of users and a distraction for the rest
- from the developer perspective: supporting it is significant effort in terms of implementing, testing and documenting. Half-ass support for some regions is not nice. Note that even after years, there's still quite a lot to do for regions, IMO if anyone has time to hack on it , it should be on core functionality, not shapely.
- If shapely becomes popular among astronomers in the future, adding support like that is easy. But we should never take things away, so if we add it now, we should support it forever.

PR #171, opened 2018-03-28
Remove out of date DS9 docs

migueldvb:
It seems that this text was accidentally added in 170d32feec3f2e86a8ea22b33536d3434aced63e.


keflavich:
is that definitely out of date?

migueldvb:
The functionality has been removed in d85a141c32005ae028817f11c069baf0c4d9b481 with the ds9 parser refactoring. It looks like this could be a problem from merging #130 at the same time that 170d32feec3f2e86a8ea22b33536d3434aced63e was introduced.

@joleroi could you confirm?

PR #172, opened 2018-04-02
Handling dimension and broadcast of x and y in PixCoord.

sushobhana:
1. converting 0d numpy arrays to float.
2. Checking whether x and y have same dimensions.
3. making more SkyCoord like : `p = PixCoord(1, [1, 2])` is valid and p.x becomes [1, 1]
4. added tests for above.

### DIFF
--- a/regions/core/pixcoord.py
+++ b/regions/core/pixcoord.py
@@ -37,19 +37,11 @@ class PixCoord(object):
         self.x = self._standardize_coordinate(x)
         self.y = self._standardize_coordinate(y)
 
-        for a, b in ((x, y), (y, x)):
-            z, shape = a, np.asanyarray(a).shape
-            if shape == () or shape == (1,):
-                while np.asanyarray(z).shape != np.asanyarray(b).shape:
-                    z = np.append(z, a)
-
-                if self.x is a:
-                    self.x = z
-                else:
-                    self.y = z
-
         if np.asanyarray(self.x).shape != np.asanyarray(self.y).shape:
-            raise ValueError('Input parameters x and y must be of same dimensions')
+            try:
+                self.x, self.y = np.broadcast_arrays(self.x, self.y)
+            except ValueError:
+                raise ValueError('Input parameters x and y cannot be broadcast')
 
     @staticmethod
     def _standardize_coordinate(val):
--- a/regions/core/pixcoord.py
+++ b/regions/core/pixcoord.py
@@ -11,7 +11,7 @@ __all__ = ['PixCoord']
 class PixCoord(object):
     """Pixel coordinates.
 
-    This class can represent scalar or array pixel coordinates.
+    This class can represent scalar or array pixel of coordinates.
 
     The data members are either numbers or Numpy arrays
     (not `~astropy.units.Quantity` objects with unit "pixel").
@@ -33,6 +33,8 @@ class PixCoord(object):
     """
 
     def __init__(self, x, y):
+        if np.asanyarray(x).shape != np.asanyarray(y).shape:
+            raise ValueError('{} and {} must be of same dimensions'.format(x, y))
         self.x = self._standardize_coordinate(x)
         self.y = self._standardize_coordinate(y)
 
@@ -49,6 +51,8 @@ class PixCoord(object):
         if isinstance(val, numbers.Number):
             return val
         else:
+            if np.asanyarray(val).shape == ():
+                return float(val)
             return np.array(val)
 
     @staticmethod
@@ -90,13 +94,7 @@ class PixCoord(object):
 
     @property
     def isscalar(self):
-        """Is this pixcoord a scalar? (a bool property)"""
-        # TODO: what's the best solution to implement this?
-        # Maybe we should sub-class ShapedLikeNDArray?
-        # See https://github.com/astropy/regions/issues/108
-        # This is a temp solution that matches the bahaviour of SkyCoord
-        return np.asanyarray(self.x).shape == ()
-        # return np.isscalar(self.x) and np.isscalar(self.y)
+        return np.isscalar(self.x)
 
     def __repr__(self):
         data = dict(name=self.__class__.__name__, x=self.x, y=self.y)
--- a/regions/core/pixcoord.py
+++ b/regions/core/pixcoord.py
@@ -11,7 +11,7 @@ __all__ = ['PixCoord']
 class PixCoord(object):
     """Pixel coordinates.
 
-    This class can represent scalar or array of pixel coordinates.
+    This class can represent scalar or array pixel coordinates.
 
     The data members are either numbers or Numpy arrays
     (not `~astropy.units.Quantity` objects with unit "pixel").
--- a/regions/core/pixcoord.py
+++ b/regions/core/pixcoord.py
@@ -34,31 +34,12 @@ class PixCoord(object):
 
     def __init__(self, x, y):
 
-        self.x = self._standardize_coordinate(x)
-        self.y = self._standardize_coordinate(y)
+        x, y = np.broadcast_arrays(x, y)
 
-        if np.asanyarray(self.x).shape != np.asanyarray(self.y).shape:
-            try:
-                self.x, self.y = np.broadcast_arrays(self.x, self.y)
-            except ValueError:
-                raise ValueError('Input parameters x and y cannot be broadcast')
-
-    @staticmethod
-    def _standardize_coordinate(val):
-        """Standardize coordinate.
-
-        Output should be either:
-        * Scalar Python number
-        * Numpy array
-
-        That's it, nothing else allowed!
-        """
-        if isinstance(val, numbers.Number):
-            return val
+        if x.shape == ():
+            self.x, self.y = np.asscalar(x), np.asscalar(y)
         else:
-            if np.asanyarray(val).shape == ():
-                return float(val)
-            return np.array(val)
+            self.x, self.y = x, y
 
     @staticmethod
     def _validate(val, name, expected='any'):
--- a/regions/core/pixcoord.py
+++ b/regions/core/pixcoord.py
@@ -33,10 +33,23 @@ class PixCoord(object):
     """
 
     def __init__(self, x, y):
-        if np.asanyarray(x).shape != np.asanyarray(y).shape:
-            raise ValueError('{} and {} must be of same dimensions'.format(x, y))
+
         self.x = self._standardize_coordinate(x)
         self.y = self._standardize_coordinate(y)
+
+        for a, b in ((x, y), (y, x)):
+            z, shape = a, np.asanyarray(a).shape
+            if shape == () or shape == (1,):
+                while np.asanyarray(z).shape != np.asanyarray(b).shape:
+                    z = np.append(z, a)
+
+                if self.x is a:
+                    self.x = z
+                else:
+                    self.y = z
+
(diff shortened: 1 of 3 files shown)


keflavich:
I don't like this change.  Scalar objects are generally easier to deal with than 0D arrays.  When working with a single region with a single central coordinate, it is counterintuitive to receive a 0D array in the result.  While using 0D arrays might supply more internal code consistency, it comes at a cost to users that is unacceptable.

cdeil:
+1 to use scalar and not 0d arrays.

I see you've removed one `FIXME` where internally we call `float` because sometimes data members aren't float (presumably 0d arrays).

Is it possible with limited effort to ensure scalar on init (and later property assignment) by passing inputs through `float`? (i.e. use 64 bit Python floats for data members)
Or would that require a ton of boilerplate property code?
If so, +1 to just move more in the direction of using float, instead of changes in the other direction, towards 0d arrays.

sushobhana:
The `PixCoord` is scalar only when it is a python number or a 0D numpy array. Most of regions the check whether the center is a scalar or not. There may be certain cases where 0d numpy array can creep in. Due to this,Some of the tests for DS9 conversions seems to break for numpy versions 1.11(or <)(see the travis-ci error logs of this PR) and also counter-intuitive for users.

And since python numbers (float) are encouraged , it seems the whole issue can be fixed if we convert 
0d numpy arrays to float (explicitly) in the `__init__` of `PixCoord`. With this, the `isscalar` becomes clearer than before and `from_sky` would also return numbers(and not 0D numpy array).

@keflavich @cdeil

cdeil:
> And since python numbers (float) are encouraged , it seems the whole issue can be fixed if we convert 0d numpy arrays to float (explicitly) in the __init__ of PixCoord.

@sushobhana - I think that would be the best solution.

sushobhana:
almost forgot about this !! @keflavich @cdeil any comments??

regions/core/pixcoord.py
```diff
@@ -11,7 +11,7 @@
 class PixCoord(object):
     """Pixel coordinates.
 
-    This class can represent scalar or array pixel coordinates.
+    This class can represent scalar or array of pixel coordinates.
```
cdeil: Very minor comment: remove "of" again, I think the previous sentence was OK. If you add "of" I think it would have to be "arrays of".

regions/core/pixcoord.py
```diff
@@ -33,6 +33,8 @@ class PixCoord(object):
     """
 
     def __init__(self, x, y):
+        if np.asanyarray(x).shape != np.asanyarray(y).shape:
```
cdeil: This is allowed in SkyCoord and ideally we'd behave the same way in PixCoord:
```
>>> from astropy.coordinates import SkyCoord
>>> SkyCoord(42, [1, 2, 3], unit='deg')
<SkyCoord (ICRS): (ra, dec) in deg
    [(42., 1.), (42., 2.), (42., 3.)]>
```
Also, the way you've written the check, to create arrays in init just to check a shape is a bit inefficient, and the error message isn't great (not mentioning the shape) but printing out the full x and y array which can be huge depending on what Numpy print options the user has set. Again, ideally we should be as similar to SkyCoord as possible:
```
>>> SkyCoord([42, 43], [1, 2, 3], unit='deg')

ValueError: Input parameters lon and lat cannot be broadcast
```
Not sure what to do here.
@sushobhana - Maybe remove the check for now and open a new reminder issue to add tests / checking for PixCoord to behave like SkyCoord?
If it's not complicated to do (I have no idea), and if you have time, you could also attempt it in this PR.
But I'm -1 on putting in this check as-is; it could even break existing users.

regions/core/pixcoord.py
```diff
@@ -90,13 +94,7 @@ def _validate(val, name, expected='any'):
...
     @property
     def isscalar(self):
-        """Is this pixcoord a scalar? (a bool property)"""
-        # TODO: what's the best solution to implement this?
-        # Maybe we should sub-class ShapedLikeNDArray?
-        # See https://github.com/astropy/regions/issues/108
-        # This is a temp solution that matches the bahaviour of SkyCoord
-        return np.asanyarray(self.x).shape == ()
-        # return np.isscalar(self.x) and np.isscalar(self.y)
+        return np.isscalar(self.x)
```
cdeil: Do we still match the behaviour of SkyCoord? Do we have any tests for that? If yes, OK.

regions/core/pixcoord.py
sushobhana: The first case that you mentioned will be easy to deal with.

The second case with the ValueError message can be modified to mention only the dimensions and not the values of x and y.

If these two things are only concern , I think it can be resolved in this PR itself.
Or else if you feel there is more to do , we can create a separate issue to deal with this later.

cdeil: @sushobhana - Then please try to resolve this in this PR. Let me know when you made a change and I should have another look.

sushobhana:
@cdeil  kindly see if it is worth the attempt!

regions/core/pixcoord.py
sushobhana: I think so. I have added some tests for this now.

regions/core/pixcoord.py
```diff
@@ -33,9 +33,24 @@ class PixCoord(object):
     """
 
     def __init__(self, x, y):
+
         self.x = self._standardize_coordinate(x)
         self.y = self._standardize_coordinate(y)
 
+        for a, b in ((x, y), (y, x)):
```
cdeil: @sushobhana - Can we avoid these ~ 10 complex lines by re-using some utility function from Astropy or Numpy?
E.g. would this help? https://docs.scipy.org/doc/numpy/reference/generated/numpy.broadcast_arrays.html
I had a quick look at SkyCoord but couldn't find the code where this input array broadcasting is done there.

regions/core/pixcoord.py
cdeil: The tests look good. Thanks!

cdeil:
@sushobhana - Please change PR title to reflect the change it does now.

regions/core/pixcoord.py
cdeil: Another option might be to always go to arrays in the init, and only handle the special case of scalars at the end of init. See https://mail.python.org/pipermail/numpy-discussion/2017-December/077512.html

Note that my previous comment on inefficiency of creating Numpy arrays was only because you just did that to get the shape, and then later again; and overall it probably wasn't the most important point / best comment.

I really like the changes you've made here, with the behaviour demonstrated via the tests, so let's try to simplify the init if possible, but in any case merge this in soon.

sushobhana: @cdeil great!!
The `numpy.broadcast_arrays` seems to work (tried in the repl), this would simplify the problem.
it is also used for  SkyCoord. The code is present in astropy/coordinates/representation.py file. (line:171)

regions/core/pixcoord.py
```diff
@@ -37,19 +37,11 @@ def __init__(self, x, y):
...
-            if shape == () or shape == (1,):
-                while np.asanyarray(z).shape != np.asanyarray(b).shape:
-                    z = np.append(z, a)
-
-                if self.x is a:
-                    self.x = z
-                else:
-                    self.y = z
-
         if np.asanyarray(self.x).shape != np.asanyarray(self.y).shape:
```
cdeil: You still have these `np.asanyarray` calls which will create two Numpy arrays unnecessarily in the scalar case.
Can't you just move the
```
x, y = np.broadcast_arrays(x, y)
```
at the top and get rid of this if and try except?

sushobhana:
@cdeil now it is short, simple and seems to solve all the problem and behaves exactly the same as skyCoord !! What do you say?

regions/core/pixcoord.py
sushobhana: Then the scalars will also be converted to numpy arrays, which we do not want.

sushobhana: We can get rid of the try except if you are ok with default ValueError exception of `numpy.broadcast_arrays`

cdeil: Yes, I would suggest to just let the ValueError from Numpy go to the user, not catch and re-throw another ValueError. And in the test I would just put this:
```
    with pytest.raises(ValueError):
        PixCoord(np.array([1, 2]), [3, 4, 5, 6])
```
(the exception message could change in future Numpy versions, and is not terribly important, so best to not assert on that message string).

cdeil: I locally tried removing the if and putting the broadcast at the front:
```
     def __init__(self, x, y):
-
+        x, y = np.broadcast_arrays(x, y)
         self.x = self._standardize_coordinate(x)
         self.y = self._standardize_coordinate(y)
 
-        if np.asanyarray(self.x).shape != np.asanyarray(self.y).shape:
-            try:
-                self.x, self.y = np.broadcast_arrays(self.x, self.y)
-            except ValueError:
-                raise ValueError('Input parameters x and y cannot be broadcast')
-
```
The only fails that occur are in repr / str because that leads to floats for int inputs, where before apparently we preserved int:
```
E       AssertionError: assert 'Region:Point...(x=3.0,y=4.0)' == 'Region:PointP...oord(x=3,y=4)'
E           Region:PointPixelRegion
E         - center:PixCoord(x=3.0,y=4.0)
E         ?                    --    --
E         + center:PixCoord(x=3,y=4)
```
Hmm ... not sure what to do.

But it's a minor point, this PR is basically ready to go.

cdeil: Looks like SkyCoord also always goes to float64 for int inputs?

cdeil: So if SkyCoord really also always goes to float64:
```
In [19]: SkyCoord(np.array([3, 4]), np.array([4, 5]), unit='deg').data.lon.dtype
Out[19]: dtype('float64')
```
then I'd suggest to do the same for PixCoord here and to update the ~ 5 assert statements on repr / str tests to pass.

@sushobhana - What do you think?

sushobhana: seems like SkyCoord always goes for float.

but the above case gets converted to float because when you broadcast scalar values , it returns 0d numpy arrays, which gets converted to float in `_standardize_coordinate()`

sushobhana: If you want every scalar to get converted to float, then yes we can change the test cases for sure. I am cool with whatever you think!

cdeil: Just found this:
https://docs.scipy.org/doc/numpy/reference/generated/numpy.asscalar.html

So how about this for init?
```
x, y = np.broadcast_arrays(x, y)
if np.isscalar(x):
    x, y = np.asscalar(x), np.asscalar(y)
self.x = x
self.y = y
```
Of if we prefer to always make float, then this:
```
x, y = np.broadcast_arrays(x, y)
if np.isscalar(x):
    x, y = float(x), float(y)
self.x = x
self.y = y
```
On that second point I have no opinion, but I still think putting broadcast_arrays first is the simplest way to write this, no?

sushobhana: `np.isscalar(np.array(2))` returns false. but if we change that to `x.shape == ()` for the first case , it would work. we can get rid of the `_standardize_coordinate()` also.

cdeil: Then let's do that shape check, and then this is ready. Thanks for your patience with my comments!

cdeil:
@sushobhana - Thank you!

PR #177, opened 2018-06-04
Added Text regions

sushobhana:
This contains a  point along with text string and the it denotes the leftmost point of the  text string.
```
In [1]: from regions import *

In [2]: from matplotlib import pyplot as plt

In [3]: reg = TextPixelRegion(PixCoord(.1,.2), "hello, This is from text regions")

In [4]: reg.plot()
Out[4]: <matplotlib.axes._subplots.AxesSubplot at 0x7f8759690f98>

In [5]: plt.show()

In [6]: reg
Out[6]: <TextPixelRegion(PixCoord(x=0.1, y=0.2), Text=hello, This is from text regions)>
```
text

For now, I have made it in `text.py`


keflavich:
Looks good! Add some tests.

regions/shapes/point.py
```diff
@@ -69,16 +68,66 @@ class PointSkyRegion(SkyRegion):
     """
 
     def __init__(self, center, meta=None, visual=None):
-        # TODO: test that center is a 0D SkyCoord
-        self.center = center
+        if center.isscalar:
+            self.center = center
+        else:
+            raise ValueError('the centre should be a 0D SkyCoord object')
```
keflavich: should capitalize The, and `centre` should be `center` (for consistency with the code)

sushobhana:
I think this isn't needed , We can just store the text in the meta dict of a point region. So, closing this. What say?

astrofrog:
Would it make sense to add this in a separate ``text.py`` file?

sushobhana:
@keflavich you guys decide whether we should keep it or not. Also, there is another symbol region where instead of a text , there will be a symbol.

keflavich:
I'm in favor of including text regions; they're useful for annotating images.  They do belong in their own `text.py` or `text_region.py` (since 'text' is more ambiguous than 'point', 'ellipse', etc. as to what sort of objects it implies)

astrofrog:
I think the only issue with text regions which makes them different from other regions is that they aren't 'absolute' in the sense they depend on e.g. font, so they might be different for different users. So we should make that clear in the documentation/docstring.

astrofrog:
One thing that is missing currently - horizontal alignment/vertical alignment of the text relative to the point, and rotation around the point.

sushobhana:
I think it would be Ok if we handle all these things in its `visual` attribute ? I will update the documentation accordingly.

PR #181, opened 2018-06-13
Update visual metadata in ds9 docs

migueldvb:


keflavich:
Why did you remove the font metadata, etc?  Those are present in most ds9 region files - all produced by ds9, at least - by default.

migueldvb:
I got this output from parsing the string `galactic\ncircle(42,43,3) # color=green` in the example. Should those metadata be included in the `reg_string` variable defined earlier?

keflavich:
No, I guess not; the minimal example has minimal associated metadata.

PR #185, opened 2018-06-15
making angle a quantity object in pixel regions

sushobhana:


sushobhana:
the tests will pass once this #179 merges.

sushobhana:
Closing this as I added this as a commit in #179 .

PR #186, opened 2018-06-18
Converting Shape objects to CRTF string

sushobhana:
- Now , symbol and text region are implemented as point regions with extra attributes in meta dict.
- Implemented `to_crtf` in ShapeList class that converts Shape objects to CRTF string.
- Moved `valid_symbols` to `io/crtf/core.py` so that `to_crtf` can also use it without any circular import errors.
- Added round tests to make sure that both reader and writer are working as expected.
- A small change in regex , fixed an overhead assertion so that it doesn't confuse between `ann` (annotation type) and  'ann' in `annulus`.
- Added a new meta key `type` which holds whether a CRTF region is  `reg` (region) or `ann` (annotation). Default is `reg`.


sushobhana:
@keflavich I think you have missed this PR, please go through this whenever you are free.

keflavich:
At a glance, this looks good.  I would like more tests, but we can add those later.

PR #187, opened 2018-06-19
Radius to width conversion for ellipse for DS9 and CRTF reader

sushobhana:
tried to solve #143 .

### DIFF
--- a/regions/io/core.py
+++ b/regions/io/core.py
@@ -172,6 +172,11 @@ class ShapeList(list):
                 temp = [val.format(x) for x in coord]
                 coord = ",".join(temp)
                 line = ds9_strings['polygon'].format(include, coord)
+            elif shape.region_type == 'ellipse':
+                coord[2:] = [x / 2 for x in coord[2:]]
+                if len(coord) % 2 == 1:
+                    coord[-1] *= 2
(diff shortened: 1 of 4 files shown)


keflavich:
These changes look like they might be fine, but describe the logic in comments so we understand it.  The `var[-1] *= 2` is for the angle, right?  The rest are for converting full-width into half-width or vice-versa.

sushobhana:
yes, I saw that in elliptical annulus angle is optional, so checking odd no. of elements in coord list for angle.
I will add them in comments. The logic seems fine right?

keflavich:
I think so. This is a good one to try to test *every* corner case, though.

migueldvb:
If the multiplication is only needed for the radius values why not specifying [2:4] ?

```                               
              self.coord[2:4] = [x * 2 for x in self.coord[2:4]] 
```

sushobhana:
Because ellipse can also represent an elliptical annulus and the no. of radius pairs are unknown.

migueldvb:
Ah OK. We could have a test for an ellipse annulus region in `test_shape_object.py`?

sushobhana:
@keflavich  I think this is good to go for now.

PR #190, opened 2018-06-22
Fix mask bug for regions with negative indices

larrybradley:
This PR fixes a bug in region mask cutouts that is triggered when the region `x` and `y` positions are both negative *and* the region has no overlap with the data.  Simple example:

```python
from regions import CirclePixelRegion, PixCoord
reg = CirclePixelRegion(PixCoord(-20, -20), radius=10.)
mask  = reg.to_mask()
data = np.ones((50, 50))
cutout = mask.cutout(data)
```

The `cutout` result should be `None`, since the region has no overlap with the data.  This PR fixes that.

This PR also add a `copy` keyword to the `cutout()` method and adds unit tests for `Mask` (there were none before), including a regression test for this issue.

This issue was fixed in `photutils` in https://github.com/astropy/photutils/pull/646.

### DIFF
--- a/regions/core/mask.py
+++ b/regions/core/mask.py
@@ -183,6 +183,8 @@ class Mask(object):
         """
 
         data = np.asanyarray(data)
+        if data.ndim != 2:
+            raise ValueError('data must be a 2D array.')
 
         partial_overlap = False
         if self.bbox.ixmin < 0 or self.bbox.iymin < 0:
--- a/regions/core/mask.py
+++ b/regions/core/mask.py
@@ -1,7 +1,10 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
-from __future__ import absolute_import, division, print_function, unicode_literals
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
+
 import numpy as np
-from astropy import units as u
+import astropy.units as u
+
 
 __all__ = ['Mask']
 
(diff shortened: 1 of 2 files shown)


sushobhana:
@larrybradley Shouldn't the `multiply` method also have the `copy` arg?

regions/core/mask.py
```diff
@@ -110,35 +135,42 @@ def to_image(self, shape):
...
+            fully within the input ``data``, then the returned cutout
+            array will be a view into the input ``data``.  In cases
+            where the mask partially overlaps or has no overlap with the
+            input ``data``, the returned cutout array will always hold a
+            copy of the input ``data`` (i.e. this keyword has no
+            effect).
+
         Returns
         -------
         result : `~numpy.ndarray`
```
sushobhana: ```result : `~numpy.ndarray` or None``` Just a suggestion.

larrybradley: Yes, thanks.  I also updated this in the `multiply` method.

larrybradley:
@sushobhana It isn't possible for the `multiply` method to return a view of the data -- it must always return a new array.  It returns an array where the mask has been multiplied by the (cutout) data.

sushobhana:
Ah! I realized , thanks for the clarification @larrybradley

larrybradley:
All passes now after rebasing (applied fix for the Travis `pytest` version for python 3.4).  Merging now.

PR #191, opened 2018-06-22
Update package infrastructure

larrybradley:
This PR updates some of the package infrastructure.  It's motivated by the fact that I couldn't run `pytest` directly within the package.  That's fixed now.


larrybradley:
Merging since this is a straightforward update from the package template.

PR #195, opened 2018-07-02
Improvements in masking for compound regions.

### LINKED ISSUE #168: CompoundPixelRegion.to_mask() cannot handle negative BoundingBoxes
As discussed in #137 the current implementation for ``CompoundPixelRegion.to_mask()`` throws an error if the bounding box extends over the borders of the array. This should be improved at some point.

sushobhana:
Fixes #168 
Handles negative entries in bounding box for compound regions.
Added tests to prove it's correctness.
Will also add a test for an annulus region.


sushobhana:
@keflavich

Issue #83, opened 2016-11-28
Attribute validation for astropy.regions?


cdeil:
In https://github.com/astropy/regions/pull/81#issuecomment-263240526 and other places in `astropy.regions` before the question came up whether we should do input validation (for `PixCoord`, `BoundingBox`, region classes with float or Angle or SkyCoord attributes, ....

I'm +1 to do it. It's some extra effort for us, but especially for interactive work it really helps users catch many little mistakes.

My preference would be to use something like [attrs](https://glyph.twistedmatrix.com/2016/08/attrs.html) or [numtraits](https://pypi.python.org/pypi/numtraits) instead of re-implementing something similar again here in `astropy.regions`.

@astrofrog @larrybradley or anyone - Thoughts?

Is there something available or proposed in Astropy core?
Should we bundle or add dependency here?
Should we discuss on the Astropy mailing list or issue tracker now, or just go ahead and start using something here and discuss later / hope they accept it in Astropy core?

cdeil:
I don't know if there's some data validation descriptors in Astropy already and given that this package is small and our needs are modest I wouldn't suggest something like attrs or traits, so having a few custom descriptor classes here in the regions package seems fine.

I had a very quick look at your descriptor implementations.

It looks like you have a mix of using `WeakKeyDictionary` and some where you use the instance to store the value and use getattr from the descriptor. Is there a logic to that or can this be simplified / unified a little?

Personally I find it weird to have a `WeakKeyDictionary` and have the values not stored on the instances, and like the pattern to keep the values in the instance dict as usual, and only use the descriptor for validation and modify the instance dict from the descriptor. This is explained [here](http://nbviewer.jupyter.org/urls/gist.github.com/ChrisBeaumont/5758381/raw/descriptor_writeup.ipynb#Beware-unhashable-descriptor-owners). Really the only drawback is that you have to put the same attribute and descriptor label name, which I don't find hard to get right because it's on a single line.

sushobhana:
Actually the logic is : the key of the `WeakKeyDictionary` is an instance of the `regions.Region` class and it's value is the value of the attribute . We use weak dictionary so that the key value pair exists until the instance is alive, hence prevents memory leakage.
We store it in this way because the descriptor classes are static to it's owner class. 
The problem with this arises only when the instances are unhashable, which is not the case for `regions.Region` (according to the link, you provided).
Thanks for the link btw!!

cdeil:
[...]

PR #196, opened 2018-07-08
Add rectangle and ellipse annulus and make region immutable.

sushobhana:
- Added `Ellipse` Annulus classes
- Added `Rectangle` Annulus classes.
- Instead of using `property` descriptors , I have created some custom descriptors for `attributes`  so that the validation method would be same for all .
Also, we need not make them forcibly `immutable`. The dependency between attributes can also be checked. For Ex: the `inner_radius` must be less than `outer_radius` of `CirclePixelRegion` 
Also. handles memory leak by using `WeakKeyDictionary`.
See #83


regions/core/core.py
```diff
@@ -38,8 +38,9 @@ def __repr__(self):
         else:
             params = []
         if self._repr_params is not None:
-            for key, val in self._repr_params:
-                params.append('{0}={1}'.format(key, val))
+            for key in self._repr_params:
+                params.append('{0}={1}'.format(key.replace("_", " "),
+                                               self.__getattribute__(key)))
```
keflavich: Might `get(self, key)` be better?

regions/shapes/annulus.py
```diff
@@ -112,3 +111,329 @@ def to_pixel(self, wcs):
...
+        _, scale, north_angle = skycoord_to_pixel_scale_angle(center, wcs)
+
+        inner_width = self.inner_width / scale * u.deg
+        inner_height = self.inner_height / scale * u.deg
+        outer_width = self.outer_width / scale * u.deg
+        outer_height = self.outer_height / scale * u.deg
+        angle = self.angle - (north_angle - 90 * u.deg)
+
+        return EllipseAnnulusSkyRegion(center, inner_width, inner_height,
+                                      outer_width, outer_height, angle,  self.meta, self.visual)
```
keflavich: inconsistent indentation here

regions/shapes/annulus.py
```diff
@@ -112,3 +111,329 @@ def to_pixel(self, wcs):
...
+
+        super(RectangleAnnulusPixelRegion, self).__init__(
+            region1=region1, region2=region2, operator=operator.xor, meta=meta, visual=visual)
+
+        self._repr_params = ('inner_width', 'inner_height',
+                             'outer_width', 'outer_height')
+
+    @property
+    def center(self):
+        return self.region1.center
```
keflavich: This is the same as for the ellipticalannulus, right?  Can we abstract them out to a metaclass?

sushobhana: I also thought about this , but then again I thought why complicate the hierarchy for the sake of a few lines since that would be illogical, IMHO.

sushobhana:
@keflavich Thanks. I will probably add some few basic test too, then we are ready to go.

regions/shapes/annulus.py
keflavich: It's not really a few lines, more like ~20-40 per region type, and it's repeated code.  I don't think it complicates the hierarchy much either.  I'm much more concerned about repeated code than about complicated inheritance, especially since the set of regions we have now may not be complete.

sushobhana: Hmm.. you are right.. But, the elliptical annulus and rectangular annulus share a lot of repeated code but not circular annulus, how do you suggest they have a meta class? 
Or may we can have a `EllipseAnnulusRegion` parent meta class for `EllipseAnnulusPixelRegion` and `EllipseAnnulusSkyRegion`?
What do you say?

keflavich: How about `AsymmetricAnnulus` as the parent class for both rectangle & ellipse annulus types?  That would capture width, height, center, angle, area for both, right?  Thought, area would have to be overloaded.

sushobhana: Seems nice. let's try that!

regions/core/core.py
```diff
@@ -293,6 +303,205 @@ def to_pixel(self, wcs):
...
+        raise NotImplementedError
+
+
+class CenterPix(RegionAttr):
+
+    def _validate(self, value):
+        if isinstance(value, PixCoord) and value.isscalar:
+            return
+        else:
+            raise ValueError('The center must be a 0D PixCoord object')
```
keflavich: might this be simpler as an `if not(condition): raise`, skipping the return statement above?

regions/core/core.py
```diff
@@ -293,6 +303,205 @@ def to_pixel(self, wcs):
...
+                             .format(self._name))
+
+
+class CenterSky(RegionAttr):
+
+    def _validate(self, value):
+        if isinstance(value, SkyCoord) and value.isscalar:
+            return
+        else:
+            raise ValueError('The center must be a 0D SkyCoord object')
```
keflavich: ditto above comment

keflavich:
You also need to fix the doctests that are failing

sushobhana:
@keflavich Ready for final review and merge.

PR #205, opened 2018-08-13
The 'contains' method is supposed to work on arrays

keflavich:
This PR fixes the broken functionality because a non-array-compatible logical not was being used.  The not operator (`~`) only works on arrays, the `not` operator only works on scalars, but `np.logical_not` appears to work on both.

@sushobhana, could you review this?  It would be helpful to have more explicit tests of both the include and not-include versions of the contains methods

### DIFF
--- a/regions/shapes/circle.py
+++ b/regions/shapes/circle.py
@@ -74,7 +74,7 @@ class CirclePixelRegion(PixelRegion):
     def contains(self, pixcoord):
         pixcoord = PixCoord._validate(pixcoord, name='pixcoord')
         in_circle = self.center.separation(pixcoord) < self.radius
-        if self.meta.get('include', False):
+        if self.meta.get('include', True):
             return in_circle
         else:
             return np.logical_not(in_circle)
--- a/regions/shapes/circle.py
+++ b/regions/shapes/circle.py
@@ -1,6 +1,7 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 from __future__ import absolute_import, division, print_function, unicode_literals
 
+import numpy as np
 import math
 
 from astropy.coordinates import Angle, SkyCoord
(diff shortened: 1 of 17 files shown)


regions/shapes/tests/test_rectangle.py
```diff
@@ -108,3 +108,8 @@ class TestRectangleSkyRegion(BaseTestSkyRegion):
         expected_str = ('Region: RectangleSkyRegion\ncenter: <SkyCoord '
                    '(ICRS): (ra, dec) in deg\n    ( 3.,  4.)>\nwidth: '
                    '4.0 deg\nheight: 3.0 deg\nangle: 5.0 deg')
+
+    def test_contains(self):
+        position = SkyCoord([1, 2] * u.deg, [3, 4] * u.deg)
+        # 1,2 is outside, 3,4 is the center and is inside
+        assert reg.contains(position) == np.array([False,True], dtype='bool')
```
sushobhana: Is it `self.reg`, then we need to add a wcs too?

sushobhana: same for all other `SkyRegion`s and pep8 space after comma?

regions/shapes/circle.py
```diff
@@ -74,9 +75,9 @@ def contains(self, pixcoord):
         pixcoord = PixCoord._validate(pixcoord, name='pixcoord')
         in_circle = self.center.separation(pixcoord) < self.radius
         if self.meta.get('include', False):
```
sushobhana: I think the default value of `include` is `True`. We should change that.

sushobhana: for all regions

keflavich: Yes... the default value of `include` is `True`, which means `self.meta.get('include', False)` is probably redundant?  But otherwise, I don't understand what you're saying.  What change are you suggesting?

keflavich: Oh, right, nevermind - if `include` is not specified, it should be `True`.  Got it.

PR #209, opened 2018-08-21
Remove shapely converter functions, keep docs page

cdeil:
This PR removes the `to_shapely` and `from_shapely` converter methods we had on some pixel region classes (point, circle, rectangle, polygon).

I did keep and update the example on the docs page explaining that Shapely exists and that astronomers might want to use it for complex planar shape analysis tasks, especially polygon-related things like buffer, union and intersect. I think the number of users that want to do that is close to zero, if someone has such analysis tasks they would likely not create Astropy region objects and convert them to Shapely objects directly, but use the direct methods to create Shapely objects via Python API or one of the serialisation formats they support. I'm actually +1 to remove that docs page also, but I gather that keeping it might increase the chance to get this change / PR accepted.

I made this suggestion in  https://github.com/astropy/regions/pull/165#issuecomment-412837404, but to have discussion in one place I'll paste my arguments here:

- `astropy-regions` should become `astropy.regions` in the core package soon.
- from the user perspective: very few astronomers need or have ever heard of Shapely, support for it doesn't belong in Astropy core IMO, it's a feature for 1% of users and a distraction for the rest
- from the developer perspective: supporting it is significant effort in terms of implementing, testing and documenting. Half-ass support for some regions is not nice. Note that even after years, there's still quite a lot to do for regions, IMO if anyone has time to hack on it , it should be on core functionality, not shapely.
- If shapely becomes popular among astronomers in the future, adding support like that is easy. But we should never take things away, so if we add it now, we should support it forever.

#46 from two years ago is on the same topic, when we started this package we weren't sure if we should use / support Shapely.

@astrofrog @keflavich @larrybradley @sushobhana - Thoughts?

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -15,7 +15,6 @@ The regions package requires the following packages:
 In addition, the following packages are needed for optional functionality:
 
 * `Matplotlib <https://matplotlib.org>`__ 1.5 or later
-* `Shapely <http://toblerity.org/shapely/manual.html>`__
 
 Stable version
 ==============
--- a/docs/shapely.rst
+++ b/docs/shapely.rst
@@ -1,45 +1,43 @@
 .. _gs-shapely:
 
-Exporting regions to shapely objects
-====================================
+Converting regions to shapely objects
+=====================================
 
 The `Shapely <http://toblerity.org/shapely/manual.html>`__ Python package is a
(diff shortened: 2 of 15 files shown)


cdeil:
@astrofrog - Please comment.

cdeil:
@keflavich @astrofrog @larrybradley - Thank you for the review and for agreeing to this change.

PR #213, opened 2018-08-29
Rectangles -> Polygons

keflavich:
It is helpful to be able to convert between these types since some things that support polygons don't support rotated rectangles

### DIFF
--- a/regions/shapes/rectangle.py
+++ b/regions/shapes/rectangle.py
@@ -11,7 +11,7 @@ from ..core import PixCoord, PixelRegion, SkyRegion, RegionMask, BoundingBox
 from .._geometry import rectangular_overlap_grid
 from .._utils.wcs_helpers import skycoord_to_pixel_scale_angle
 from ..core.attributes import ScalarPix, ScalarLength, QuantityLength, ScalarSky
-from .polygon import PixelPolygonRegion
+from .polygon import PolygonPixelRegion
 
 
 __all__ = ['RectanglePixelRegion', 'RectangleSkyRegion']
--- a/regions/shapes/rectangle.py
+++ b/regions/shapes/rectangle.py
@@ -203,27 +203,15 @@ class RectanglePixelRegion(PixelRegion):
         Return the x, y coordinate pairs that define the corners
         """
 
-        radius = ((self.width / 2.)**2 + (self.height/2.)**2)**0.5
-        ang2 = np.arctan2(self.width, self.height)*u.rad
-        sint = np.sin(self.angle + ang2)
-        cost = np.cos(self.angle + ang2)
-        dd1 = np.abs(radius * sint)
-        dd2 = np.abs(radius * cost)
+        corners = np.array([(self.center.x-self.width/2, self.center.y-self.height/2),
+                            (self.center.x+self.width/2, self.center.y-self.height/2),
+                            (self.center.x+self.width/2, self.center.y+self.height/2),
+                            (self.center.x-self.width/2, self.center.y+self.height/2),
+                           ])
+        rotmat = np.array([[np.cos(self.angle), np.sin(self.angle)],
+                           [-np.sin(self.angle), np.cos(self.angle)])
 
-        if (self.angle % 180*u.deg) < 90*u.deg and self.height > self.width:
-            corners = [(self.center.x - dd1, self.center.y - dd2),
-                       (self.center.x - dd2, self.center.y - dd1),
-                       (self.center.x + dd2, self.center.y + dd1),
-                       (self.center.x + dd1, self.center.y + dd2),
-                      ]
-        else:
-            corners = [(self.center.x - dd1, self.center.y + dd2),
-                       (self.center.x - dd2, self.center.y + dd1),
-                       (self.center.x + dd2, self.center.y - dd1),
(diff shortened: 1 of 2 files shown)


regions/shapes/rectangle.py
```diff
@@ -197,6 +197,23 @@ def as_patch(self, origin=(0, 0), **kwargs):
         return Rectangle(xy=xy, width=width, height=height,
                          angle=angle, **mpl_params)
 
+    @property
+    def corners(self):
+        """
+        Return the x, y coordinate pairs that define the corners
+        """
+
+        corners = np.array([(-self.width/2, -self.height/2),
```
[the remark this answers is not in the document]
keflavich: :+1:, this is just instinct.

keflavich:
Added the test.  I really hope it's right. 

Also added a `to_poly` converter.  It would be nice to have these for all regions, even circles, with `nverts` as a free parameter in that case

regions/shapes/tests/test_rectangle.py
```diff
@@ -25,6 +25,21 @@ def wcs():
...
+
+    xc,yc = 2,2
+    angle = 30*u.deg
+    width = 2
+    height = 1
+    reg = RectanglePixelRegion(center=PixCoord(xc, yc),
+                               width=width, height=height, angle=angle)
+
+    y1 = yc + np.cos(angle) * height/2 + np.sin(angle) * width/2
+    x1 = xc + np.cos(angle) * width/2 - np.sin(angle) * height/2
```
cdeil: Personally I don't like tests where the method is re-implemented. It's likely that if you make a mistake with the rotation matrix in the code, you'd make the same mistake re-coding it in the test.

Suggest to just apply the method and assert_allclose on the numbers that come out.

If you can pick a test case where you either know exactly (e.g. 90 deg rotation) or kinda (e.g. 1 deg rotation) what should come out, then put that, but otherwise IMO also just arbitrary output numbers that establish the current behaviour are better than re-implementing the method.

But I know many people disagree and prefer to write tests this way, so no need to argue. Put as you think best and 👍 to merge.

regions/shapes/rectangle.py
```diff
@@ -197,6 +198,31 @@ def as_patch(self, origin=(0, 0), **kwargs):
...
+                   ( self.width/2,  self.height/2),
+                   (-self.width/2,  self.height/2),
+                  ]
+        rotmat = [[np.cos(self.angle), np.sin(self.angle)],
+                  [-np.sin(self.angle), np.cos(self.angle)]]
+
+        return np.dot(corners, rotmat) + np.array([self.center.x,
+                                                   self.center.y])
+
+    def to_polygon(self):
```
cdeil: Add test that executes to_polygon once?

cdeil:
>  It would be nice to have these for all regions, even circles, with nverts as a free parameter in that case

Yes, that can be nice. Open a feature request issue?

I think both matplotlib and shapely (and surely others) have such code already, so whoever does this more systematically should probably look there for guidance instead of starting from scratch. (some shapes aren't easy to polygonise)

regions/shapes/tests/test_rectangle.py
```diff
@@ -25,6 +25,36 @@ def wcs():
...
+
+    assert (x1, y1) in reg.corners
+
+    reg = RectanglePixelRegion(center=PixCoord(xc, yc),
+                               width=width, height=height, angle=90*u.deg)
+    # simple case: rotate by 90
+    assert {(2.5, 3), (1.5, 3), (2.5, 1), (1.5,1)} == set(map(tuple,
```
cdeil: Why use a Python set? Just assert on the list or array that comes out, then you also check the order of corners, i.e. establish exactly what the method does.

keflavich: laziness.  Which is weird because it was a lot more text the wrong way.

What I really wanted to do was make a test that was... iso...something?  isomorphic? under rotation, e.g., such that:
[1,2,3,4] == [2,3,4,1] != [1,3,2,4]
since technically that's what matters for a polygon, but I never figured out the appropriate term and moved on...

regions/shapes/tests/test_rectangle.py
keflavich: I decided I want both kinds of test, so now both exist.  In particular, I wanted at least one case that was not just 90 deg rotations, since the intermediate rotations (and also not 45 deg rotations...) are somewhat nontrivial

regions/shapes/rectangle.py
```diff
@@ -197,6 +198,33 @@ def as_patch(self, origin=(0, 0), **kwargs):
...
+        """
+        Return a 4-cornered polygon equivalent to this rectangle
+        """
+        vertices = PixCoord(x=[c[0] for c in self.corners],
```
cdeil: You are accessing self.corners and thus computing that twice. Does this work?
```
x, y = self.corners
```
Or maybe some other Numpy array expression? But computing self.corners twice and doing Python list comprehensions seems sub-optimal.

keflavich: Completely agree that it is suboptimal.  I'll do it a better way...

keflavich:
thanks for the pep8 check; it revealed that my syntax checker was broken

Issue #215, opened 2018-08-29
Several plotting problems


keflavich:
Some issues with plotting:
 * All points are shown as circles by default.  They should be configurable instead.
 * The point circles fail to display in the specified color (unclear why)
 * the `origin` keyword does nothing in `as_patch` and `plot` methods

There might be others.

sushobhana:
> All points are shown as circles by default. They should be configurable instead.

They are configurable.
```
In [3]: from regions import *

In [4]: reg = PointPixelRegion(PixCoord(0.5,.5))

In [5]: reg
Out[5]: <PointPixelRegion(PixCoord(x=0.5, y=0.5))>

In [6]: reg.visual['symbol'] = 'd' # matplotlib symbols

In [7]: reg.plot()
Out[7]: <matplotlib.axes._subplots.AxesSubplot at 0x7fefd9645cc0>

```
![e1](https://user-images.githubusercontent.com/22423652/44802965-b86a6200-abda-11e8-9147-884c3d4e2c99.png)

> The point circles fail to display in the specified color (unclear why)
```
In [10]: reg.visual['color'] = 'red'

In [11]: reg.plot()
Out[11]: <matplotlib.axes._subplots.AxesSubplot at 0x7fefd9386ba8>
```
![e2](https://user-images.githubusercontent.com/22423652/44803048-e9e32d80-abda-11e8-8040-6d7f9d461210.png)

> the origin keyword does nothing in as_patch and plot methods

any particular region?? Because I have shifted the centers for many (almost all) regions.

keflavich:
Note that `reg.plot` allows configurable plot symbols, but `as_patch` does not.  It needs to.  

And, yes, point regions don't respect origin

cdeil:
@keflavich - Please put a milestone: is this v0.3 (resolved next week latest) or v0.4?

sushobhana:
> Note that reg.plot allows configurable plot symbols, but as_patch does not. It needs to.

There are no matplotlib patches for various symbols, so we should not be using `as_patch` method for points. (It should be probably be removed)

> And, yes, point regions don't respect origin

sorry! I probably must have missed this.

keflavich:
Milestoned for v0.3 because I believe #218 solves this... but clearly we don't have good test coverage...

PR #216, opened 2018-08-29
origin offset in `plot` and `as_patch` of PointPixelRegion

sushobhana:
It is implicit that the `as_patch` of the point region gives a circle patch 
which I think should be deprecated to avoid confusion.

The `plot` draws a `Line2D` which actually represents a point.
#215 
@keflavich

### DIFF
--- a/regions/core/core.py
+++ b/regions/core/core.py
@@ -246,7 +246,7 @@ class PixelRegion(Region):
             The ``(x, y)`` pixel position of the origin of the displayed image.
             Default is (0, 0).
 
-        kwargs: `dict`
+        kwargs : `dict`
             keywords that a `~matplotlib.patches.Patch` accepts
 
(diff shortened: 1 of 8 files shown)


keflavich:
:+1:, though I might have also independently done this in #218.

It looks like there might be an incorrect parameter order in some plot call somewhere though?
https://travis-ci.org/astropy/regions/jobs/422233806#L1322
e.g., plot(ax, origin) instead of plot(origin, ax)....

cdeil:
@sushobhana - Is this PR still relevant? If yes, please resolve the merge conflict. If no, please close.

sushobhana:
@cdeil - sorry for not coming to it sooner. looking into it now.

sushobhana:
@keflavich - ready to go??

PR #218, opened 2018-08-29
Refactor as_patch->as_artist

### LINKED ISSUE #215: Several plotting problems

keflavich:
Convert `as_patch` to `as_artist` and add a generic `plot` method for pixel regions.

Fixes #215

### DIFF
--- a/regions/core/core.py
+++ b/regions/core/core.py
@@ -236,7 +236,7 @@ class PixelRegion(Region):
                                  " a strictly positive integer)".format(subpixels))
 
     @abc.abstractmethod
-    def as_patch(self, origin=(0, 0), **kwargs):
+    def as_artist(self, origin=(0, 0), **kwargs):
         """
         Convert to mpl patch
 
@@ -290,7 +290,7 @@ class PixelRegion(Region):
 
     def plot(self, origin=(0, 0), ax=None, **kwargs):
         """
-        Calls ``as_patch`` method forwarding all kwargs and adds patch
+        Calls ``as_artist`` method forwarding all kwargs and adds patch
         to given axis.
 
         Parameters
@@ -313,7 +313,7 @@ class PixelRegion(Region):
         if ax is None:
             ax = plt.gca()
 
-        patch = self.as_patch(origin=origin, **kwargs)
+        patch = self.as_artist(origin=origin, **kwargs)
         ax.add_patch(patch)
 
         return ax
--- a/regions/core/core.py
+++ b/regions/core/core.py
@@ -256,25 +256,6 @@ class PixelRegion(Region):
         """
         raise NotImplementedError
 
-    def plot(self, ax=None, origin=(0, 0), **kwargs):
-        """
-        Create an artist and add it to the specified plot
-
-        Parameters
-        ----------
-        origin : array_like, optional
-            The ``(x, y)`` pixel position of the origin of the displayed image.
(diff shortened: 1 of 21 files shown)


cdeil:
In Gammapy, we use regions and call as_patch once:
https://github.com/gammapy/gammapy/search?q=as_patch&unscoped_q=as_patch
That's OK, we can adapt, but we shouldn't change back and forth and cause churn.

So @astrofrog @larrybradley @sushobhana  - do you agree `as_artist` is the better method name than `as_patch` here?

keflavich:
Agreed, we should not go back on this.

We *could* have both `as_patch` and `as_artist` to allow access to patches when patches are the appropriate type.  I don't know if it would be useful, but it would be relatively harmless and would prevent backward compatibility breaking

cdeil:
Suggest to choose either `as_patch` or `as_artist`. There's very few users now, but there will be in the next 10+ years for `astropy.regions`.

So `as_artist` is accurate? All objects returned are MPL Artist objects?

keflavich:
Yes, should be.  Would be good to test this....

keflavich:
Yes, will complete that one first.

regions/core/core.py
```diff
@@ -256,6 +256,26 @@ def as_patch(self, origin=(0, 0), **kwargs):
         """
         raise NotImplementedError
 
+    def plot(self, ax=None, origin=(0, 0), **kwargs):
```
sushobhana: This method is already present.

regions/shapes/text.py
```diff
@@ -58,7 +58,7 @@ def to_sky(self, wcs):
         center = pixel_to_skycoord(self.center.x, self.center.y, wcs=wcs)
         return TextSkyRegion(center, self.text)
 
-    def as_patch(self, **kwargs):
+    def as_artist(self, **kwargs):
```
sushobhana: we should implement this, then the overidden `plot` method can be removed.

regions/shapes/point.py
```diff
@@ -105,15 +105,22 @@ def as_patch(self, origin=(0, 0), **kwargs):
...
         mpl_params.update(kwargs)
+        point = Line2D([self.center.x + origin[0]],
+                       [self.center.y + origin[1]],
+                       **mpl_params)
 
-        return Circle((self.center.x, self.center.y), radius=2, **mpl_params)
+        return point
 
-    def plot(self, origin=(0, 0), ax=None, **kwargs):
+    def plot(self, ax=None, origin=(0, 0), **kwargs):
```
sushobhana: This should be removed.

regions/core/core.py
keflavich: good catch

regions/shapes/point.py
keflavich: indeed, it's now redundant with the core plot method

keflavich:
I think we should change the `plot` method to take `axis` as the first argument instead of `origin` being the first argument.  I also think it should be `axis`, not `ax`.  Any objections to me making this change?  Making it here?

keflavich:
I still suggest we change the plot arguments, but let's do that in another PR.

PR #220, opened 2018-09-08
fixed test locally

sushobhana:
A test that was failing locally is fixed now. Arised due to #218


sushobhana:
We will proceed the release once this is merged

PR #226, opened 2018-11-16
Change CRTF writer to match CASA implementation

keflavich:
CASA is incompatible with its own specification.  It cannot read CRTF files with 'global' coordinate specifications, even though the documented spec explicitly shows that example.  I've made a few fixes that empirically solve _some_ of the issues with the CASA parser, but frankly, CASA effectively does not have any capability of handling its own region files.  AFAICT, if a region file contains a region outside the image it's being "applied" to, it simply crashes and ignores the rest of the file.


aakepley:
@keflavich -- You've probably already checked this, but is CARTA going to change this situation?

PR #227, opened 2018-11-16
Support boolean array mask creation

keflavich:
This is a WIP, with only one commit so far.

A critical tool for regions is converting a region or list of regions into a boolean array mask for a given image.  AFAICT, we don't support that right now, but I might be missing something.

Note that https://astropy-regions.readthedocs.io/en/latest/masks.html does not include this use case.  It only shows how to make cutouts and cutout masks.  It does not show how to make complete boolean array masks for an image.

The first step in this process is making `to_image_partial_overlap` a public function, because that's effectively all that is needed.  You just create a blank mask array with the same shape as the target image, then use this function to fill in the blanks.

The goal of this PR is to produce something with functionality identical to pyregion.get_mask.

### DIFF
--- a/regions/core/mask.py
+++ b/regions/core/mask.py
@@ -99,10 +99,17 @@ class RegionMask(object):
 
         return slices_large, slices_small
 
-    def _to_image_partial_overlap(self, image):
+    def to_image_partial_overlap(self, image):
         """
-        Return an image of the mask in a 2D array, where the mask
-        is not fully within the image (i.e. partial or no overlap).
+        Return an image of the mask in a 2D array, where the mask is not fully
+        within the image (i.e. partial or no overlap).  This can be used, for
+        example, to convert a region to a boolean array mask of an image.
+
+        Parameters
+        ----------
+        image : np.array
+            An ndarray image whose values will be modified inplace to be the
+            values of the mask.
         """
 
         # find the overlap of the mask on the output image shape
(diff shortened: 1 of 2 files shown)


regions/core/mask.py
```diff
@@ -99,10 +99,17 @@ def _overlap_slices(self, shape):
 
         return slices_large, slices_small
 
-    def _to_image_partial_overlap(self, image):
+    def to_image_partial_overlap(self, image):
```
nitish705702: def _to_image_partial_overlap(self, image):

we do not necessary even want to change

nitish705702: but u can do.

keflavich: I don't understand your comment.

larrybradley:
@keflavich Sorry, I didn't see this issue until now.  You can create an image (of any desired shape) from the mask object.  Recall that the regions/masks don't contain any information about external data/images.  

An image of the mask can be created using the `to_image(shape)` method on the `RegionMask` object, e.g.:

```python

center = PixCoord(4., 5.)
reg = CirclePixelRegion(center, 2.3411)
mask = reg.to_mask()
image = mask.to_image((50, 50))
```

I think all that is needed is to add an example of `to_image()` in the documentation.

keflavich:
Right, that makes sense.  Could you copy your code example over to another Issue and mark it as a documentation issue (...or just fix the docs, depending on what kind of time you have...) and then close this one?  I'll leave this open until we have a replacement.

nitish705702:
Sir please help me .i want to this projct apply for gsoc.

PR #234, opened 2019-01-16
Fix for issue 233: 'text' renamed to 'label'

keflavich:
WIP: do not merge.

In a previous commit, the metadata 'text' was renamed to 'label'.  This breaks the one-to-one mapping between ds9 metadata keywords and astropy.regions metadata keywords, and therefore is a questionable decision.  We need to determine, going forward, whether to:

 * [ ] revert to using 'text' as the text metadata label
 * [ ] stick with using 'label' as the text metadata label
 * [x] keep both, so that reg.meta['text'] = reg.meta['label']

Any feedback?


keflavich:
This is ready for review / merging once.... actually once nothing happens?  I guess we need to merge #251 first, since I based this PR off that one

keflavich:
Assuming this passes, will merge.

Issue #245, opened 2019-03-10
Astropy/Region Tutorial "Computing Overlap Masks" error


Vab-jain:
In the "Computing Overlap Masks" [tutorial,](https://astropy-regions.readthedocs.io/en/latest/masks.html) the example to demonstrate different `to_mask()` modes does not show the plot due to missing `plt.show()` command.
![tut_error](https://user-images.githubusercontent.com/22465402/54084132-d74c6c80-4352-11e9-8a4b-c3de4b398fab.png)

Here is a fixed version of it.
![tut_fix](https://user-images.githubusercontent.com/22465402/54084155-2db9ab00-4353-11e9-91ff-dce295bb6d98.png)

Similar error is also there in the next example of `to_image()` method

keflavich:
This is a user settings issue.  If you set `pl.ion()`, the images will display.

PR #246, opened 2019-03-12
Update masks.rst

Vab-jain:
In response to issue #245 . Adding `plt.show()` to show resulted image in the example for `to_mask()` and `to_image()` methods.


keflavich:
The documentation renders correctly now:
https://astropy-regions.readthedocs.io/en/latest/masks.html
so this change is not required.  `pl.show` is only needed if you have `pl.ioff()` set; it is dependent on global user settings.

PR #251, opened 2019-03-26
Remove conftest.py

keflavich:
conftest.py has only one active line and that line is invalid; it uses a deprecated astropy path for pytest-plugins


keflavich:
This was dumb, I just hadn't pulled the latest changes.

PR #252, opened 2019-03-27
Replace calls to deprecated np.asscalar() by .item()

adonath:
Since numpy 1.16 the `np.asscalar` is deprecated. This PR replaces the two calls in `pixcoord.py` by the equivalent `.item()`.

### DIFF
--- a/regions/core/pixcoord.py
+++ b/regions/core/pixcoord.py
@@ -37,7 +37,7 @@ class PixCoord(object):
         x, y = np.broadcast_arrays(x, y)
 
         if x.shape == ():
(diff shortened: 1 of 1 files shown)


keflavich:
Is this backward-compatible?

adonath:
@keflavich From a quick look it seems `.item()` is already there since numpy 1.9 or earlier (https://docs.scipy.org/doc/numpy-1.9.2/reference/generated/numpy.ndarray.item.html#numpy.ndarray.item). So this change should be backwards compatible...

cdeil:
Astropy did the same edit: https://github.com/astropy/astropy/pull/7880/files

@keflavich - merge?

PR #256, opened 2019-05-09
Fix ReadTheDocs

larrybradley:
RTD hasn't been building the `regions` docs for almost 5 months now (due to the deprecated github integration).  I switched to the github webhook, but RTD now has build failures due the conda environment.  This PR switches to use pip on RTD.


larrybradley:
RTD is working again. :smile:

Issue #266, opened 2019-06-07
Add region copy method


cdeil:
We would like to add a copy method to all regions, that returns a deep and independent copy of all data members.

Currently there is no easy way to make a copy of a region as far as I know.

The first use case for region copy would be the region rotate method proposed in  #265 .
To create the rotated region, you first make a copy, then modify a few data members, then return the new object.

@astrofrog @keflavich - Thoughts?

cdeil:
Done in #269 #281 #283 #284

Issue #265, opened 2019-06-07
Add region rotate method


cdeil:
We would like to add a method to the region classes to rotate around a given point by a given angle.

This is needed in Gammapy, because in ground-based gamma-ray astronomy one common background estimation technique is to point off-source (called "wobble offsets") and to estimate the background in a given "on" or "source" region from "off" regions which are rotated around the pointing position, following the good assumption that background in the field of view is rotationally symmetric.

This is illustrated [here](https://docs.gammapy.org/0.12/background/reflected.html) and already implemented in Gammapy in the [ReflectedRegionsFinder](https://docs.gammapy.org/0.12/api/gammapy.background.ReflectedRegionsFinder.html), but it turns out that computing rotated regions for different shapes is difficult, it leads to if, else, getattr.

It would be much simpler to implement this as
```
rotated_region = region.rotate(point, angle)
```
and to have `rotated_region` be an independent new object with some attributes changed (like e.g. center and angle for a rectangle).

@astrofrog @keflavich - OK if we add this method?

The added method doesn't interfere with the rest of the regions code, and we promise to add tests and maintain it.

There are some details to be worked out, the most important question probably being whether this is added only for pixel regions (what we currently use for this in Gammapy), or in addition also for sky regions (what we might want to use in Gammapy). Another question is if all regions support this or not, i.e. if this should be a required method via the `PixRegion` ABC, or just added to the regions that do support it. These things will become clear once we start to implement it...

keflavich:
Huge :+1: from me.  I think we need this functionality if and when we make a region editor gui

cdeil:
Just to be clear: there is an option to implement this feature by adding a `rotation_center` and `rotation_angle` data member to region objects, and then to apply rotations on the fly in contains and draw and other operations. I've seen this in one (non-public) gamma-ray code, and I think such a "transform" feature is also sometimes used in graphics programs.

My suggestion here was to **not** do that, but just to add a method that makes a copy, then modifies the exiting data members like `center` or `angle`, and returns the new object.

So if anyone thinks it should be implemented in a different way, please comment here.

keflavich:
I think your approach would make the GUI editing a little more difficult.  I agree generally that it's the better thing to do, but please keep in mind that we may _internally_ want editable objects.

cdeil:
Implemented in #281 and #285

PR #269, opened 2019-06-10
Add region copy methods

cdeil:
As mentioned in #266 I'd like to add copy methods for regions.

I looked around a bit, e.g. [here](https://docs.python.org/3/library/copy.html) and [here](https://pymotw.com/3/copy/index.html), and there's `__copy__` and `__deepcopy__` that we could implement.

However, looking at `SkyCoord`, that doesn't seem to be implemented there. So I'd suggest to follow that class, and add a `copy` method for regions.

@astrofrog @keflavich @larrybradley  @sushobhana or anyone interested - could you please have a look at the one example method and test fr `CirclePixelRegion.copy` here? Is it OK, or would you suggest some different implementation or test pattern?

I'll wait a day here for suggestions before going ahead and adding `copy` methods for many region classes.

### DIFF
--- a/regions/core/pixcoord.py
+++ b/regions/core/pixcoord.py
@@ -1,6 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 from __future__ import absolute_import, division, print_function, unicode_literals
-import numbers
+import copy
 import numpy as np
 from astropy.coordinates import SkyCoord
 from .core import _DEFAULT_WCS_ORIGIN, _DEFAULT_WCS_MODE
@@ -40,6 +40,9 @@ class PixCoord(object):
             self.x, self.y = x.item(), y.item()
         else:
             self.x, self.y = x, y
+
+    def copy(self):
+        return self.__class__(copy.copy(self.x), copy.copy(self.y))
 
     @staticmethod
     def _validate(val, name, expected='any'):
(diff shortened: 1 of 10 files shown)


larrybradley:
LGTM, carry on!

cdeil:
@keflavich or @larrybradley - Could you please review this?

I've added copy for circle, ellipse, rectangle and polygon here.
And would prefer to continue with other shapes in future PRs.

cdeil:
We could also implement `copy` in a generic way if we have a list of data fields for each class. And we could allow changes to be made on copy, which is very commonly what people want: make a copy, but change a few fields. I'm not sure if this generic `copy` written once in the base class would work for now, for the annuli classes (see #279).

See example implementing this proposal for one class: https://github.com/astropy/regions/commit/93c36566360b1b4e2c556c59ae5192b71d1213b9

This is similar to what [replace](https://github.com/python/cpython/blob/d561f848b235f2011a43b705d112055b92fa2366/Lib/dataclasses.py#L1205-L1253) does for [dataclasses](https://docs.python.org/3/library/dataclasses.html)

cdeil:
I'll merge this in now, because I've gone ahead and started to implement `rotate` (see #265) which requires this copy functionality. I'll open a new PR for that shortly where I also change the implementation of `copy`. Comments still welcome any time!

PR #271, opened 2019-06-11
CRTF regions can have more text on line 1

keflavich:
very small correction.  TODO: needs a test!

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -54,6 +54,9 @@ Bug Fixes
 - Fixed an issue where the CRTF file parser would fail if the CRTF
   version number was included on the first line. [#240]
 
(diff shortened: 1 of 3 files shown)


keflavich:
This needs to be merged; CRTF reading is broken because of this.  But I want a test.  Hrmph.

keflavich:
Also adding a regression test: spaces are allowed (in CARTA's spec, at least) after the region type name

larrybradley (approved):
Thanks, @keflavich.  There were 3 different things going on in the original PR, one of which was already fixed by you in #240.  This PR addresses the space after the region name.  The third issue (closed polygons) was broken in this PR.  I'll open a separate PR fix.

PR #272, opened 2019-06-11
Minor improvements to mask docs

keflavich:
The mask docs are somewhat unclear at present; `mask.multiply` is not
appropriate for most statistics.  Also, I think we need a `mask.get_pixels`
method to enable use of other types of statistics, like the median.  Do we have
that somewhere already and I'm just forgetting?  Seems likely.

### DIFF
--- a/docs/masks.rst
+++ b/docs/masks.rst
@@ -163,8 +163,11 @@ as well as a cutout with the data multiplied by the mask::
     >>> weighted_data = mask.multiply(hdu.data)
     >>> pf.close()
 
+Note that ``weighted_data`` will have zeros where the mask is zero; it therefore
+should not be used to compute statistics.  See `Masked Statistics` below.
+
 We can take a look at the results to make sure the source overlaps
-with the aperture::
+with the aperture
 
 .. doctest-skip::
 
@@ -240,6 +243,9 @@ at the extent of the mask in the image:
    :nofigs:
 
     pf.close()
+
+Masked Statistics
+`````````````````
 
 Finally, we can use the mask and data values to compute weighted
 statistics::
--- a/docs/masks.rst
+++ b/docs/masks.rst
@@ -167,7 +167,7 @@ Note that ``weighted_data`` will have zeros where the mask is zero; it therefore
 should not be used to compute statistics.  See `Masked Statistics` below.
 
 We can take a look at the results to make sure the source overlaps
-with the aperture
+with the aperture:
 
 .. doctest-skip::
 


larrybradley:
@keflavich It's a bit tricky because only the `.to_mask(mode='center')` mode returns a mask with values only 0 or 1.  For that mode only will `mask.multiply` get you the original data pixels.  The other modes return fractional pixel masks, therefore one would need to use weighted statistics on the result of `mask.multiply` (or the non-zero pixels in the result of `mask.multiply`).  `np.average` (not `np.mean`) supports the use of weights, but there isn't a common set of tools for other weighted statistics.  It's been on my to-do list to add a weighted-median function to `astropy.stats`.

Perhaps we could add a keyword option to the `mask.multiply` method that would return a 1D array of weighted non-zero pixels instead of the full 2D mask-weighted array (which contains zeros), with a well-documented caveat that it's "user-beware" when not using a mask defined with `mode='center'`.  I'm not sure what to name such a keyword.  `mask.multiply(data, return_nonzero_weighted_pixels_as_1d=True)` is probably the most descriptive, but horribly long.

larrybradley:
On second thought, perhaps `mask.multiply` should be left alone and instead we add a new method called something like `mask.multiply_1d` that returns the non-zero weighted pixels as a 1D ndarray.

keflavich:
I like the `multiply_1d` approach, I think that solves a lot of the issues.  However, I'm not sure about the name.  `mask.multiply` still remains counterintuitive to me because I so rarely work with fractional pixel weights.  What about `mask.get_valid_pixels` or something along those lines? I'm open to other suggestions (and not feeling very creative at the moment), but I want a method name that better conveys what it does.

Also - is this PR OK?  I think it's just highlighting a caveat more clearly.

docs/masks.rst
```diff
@@ -163,8 +163,11 @@ as well as a cutout with the data multiplied by the mask::
     >>> weighted_data = mask.multiply(hdu.data)
     >>> pf.close()
 
+Note that ``weighted_data`` will have zeros where the mask is zero; it therefore
+should not be used to compute statistics.  See `Masked Statistics` below.
+
 We can take a look at the results to make sure the source overlaps
-with the aperture::
+with the aperture
```
larrybradley: You can keep a single ":" here.  The double "::" is what triggered the `doctest-skip::` showing up in the docs.

larrybradley:
It was @astrofrog who renamed it to `mask.multiply()` (since the result is the mask multiplied with the data).  I think I originally called it `mask.apply()`.  `mask.multiply()` makes sense to me, but I almost always use fractional pixel weights.  Thoughts, @astrofrog?

astrofrog:
I think ``get_valid_pixels`` will be ambiguous when there are fractional pixels, hence why I'd chosen ``multiply`` originally, but I'm open to other names.

larrybradley (approved):
Thanks, @keflavich.  BTW, I added a `get_values()` method in #351.

PR #278, opened 2019-06-12
Fix corner-case issue in RegionMask.multiply()

larrybradley:
This PR fixes a corner-case bug where `RegionMask.multiply()` would not set non-finite (NaN, +/- inf) data values outside of the mask but within the mask bounding box to zero.  The array returned by `multiply` should be nonzero *only* within the mask.


larrybradley:
@cdeil Done!

Issue #57, opened 2016-07-21
Polygon region vertices


larrybradley:
The `PolygonPixelRegion` is initialized with a `vertices` input, which it appears are _absolute_ positions.  That makes it a bit more difficult to move the region to different positions.  Instead, should the `vertices` define a shape relative to `(0, 0)` and then have an additional `position` (or `center`?) parameter that defines the absolute position (i.e. a shift from `(0, 0)`)?

cdeil:
I don't care either way. Having absolute as data members has the advantage that no additional center is needed. The non-public region package I have been using so far even had a rotation point and rotation angle as data member for every region, because rotations were a common operation. I'm not advocating for this here.

My suggestion: whoever volunteers to implement polygons gets to choose. :-)

How does `shapely` and `sphere` do this? Do they have extra data members for such a shift on the polygon objects?

astrofrog:
If we do include this, I would suggest calling it ``offset`` (center would be misleading and position might be confusing since there are multiple positions).

cdeil:
I'm moving the milestone to 0.3.
But if someone has time to work on polygons within the next ~week, it could still go in the 0.2 release.

@larrybradley - Could you please comment again what you think should be done for polygons, based on the discussion here? How about copying over the pixel polygon code from Ginga  (see links in #69)?

larrybradley:
@cdeil My motivation to define the polygon using *relative* positions (e.g. from zero) instead of *absolute* positions was for aperture photometry.  In other words, the *shape* of the polygon can be defined by the *relative* positions.  That would be like a `PolygonAperture` object.  The `PolygonAperture` object could then be applied to the data at different absolute positions.

If the region is defined using *absolute* vertices, I would need to redefine a new `PolygonAperture` for every new position it is to be applied to the data.

larrybradley:
BTW, if the polygon is defined with *relative* positions, we would need to add an additional `position` (`origin`?) keyword for the absolute position.

larrybradley:
@cdeil  It looks like `ginga` is using *absolute*-position vertices since the region is defined by drawing on a canvas.

cdeil:
Moved to 0.4 milestone.

cdeil:
See comment by @larrybradley in https://github.com/astropy/regions/pull/281#issuecomment-502838913

That PR is closed, let's continue the discussion here and try to find a solution.

---

[...]

PR #281, opened 2019-06-13
Change copy and add rotate methods

cdeil:
This PR is a follow-up to #269 - I did implement the suggestion from https://github.com/astropy/regions/pull/269#issuecomment-501575413 and put `copy` in the base class. I'm not sure if that's good - comments welcome!

This PR also adds PixCoord and pixel region rotate methods, as I mentioned in #265 .

I'll continue a bit and add copy and rotate for a few more regions here.

@astrofrog @keflavich @registerrier or anyone - please review.

### DIFF
--- a/regions/core/core.py
+++ b/regions/core/core.py
@@ -3,6 +3,7 @@ from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 import abc
 import six
+import copy
 import operator
 import inspect
 
@@ -65,6 +66,14 @@ class Region(object):
     """
     Base class for all regions.
     """
+
+    def copy(self, **changes):
+        """Make an independent (deep) copy."""
+        for field in self._fields:
+            if field not in changes:
+                changes[field] = copy.deepcopy(getattr(self, field))
+
+        return self.__class__(**changes)
 
     def __repr__(self):
         if hasattr(self, 'center'):
--- a/regions/core/core.py
+++ b/regions/core/core.py
@@ -69,7 +69,15 @@ class Region(object):
 
     def copy(self, **changes):
         """Make an independent (deep) copy."""
-        for field in self._fields:
+        fields = []
+        if hasattr(self, "center"):
+            fields.append("center")
+        if hasattr(self, "vertices"):
+            fields.append("vertices")
+        fields.extend(self._repr_params)
+        fields.extend(["meta", "visual"])
+
+        for field in fields:
             if field not in changes:
                 changes[field] = copy.deepcopy(getattr(self, field))
 
--- a/regions/core/pixcoord.py
+++ b/regions/core/pixcoord.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 import copy
 import numpy as np
 from astropy.coordinates import SkyCoord
+
 from .core import _DEFAULT_WCS_ORIGIN, _DEFAULT_WCS_MODE
 
 __all__ = ['PixCoord']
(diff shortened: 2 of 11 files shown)


larrybradley:
Each region already defines a `_repr_params` attribute for the shape parameters.  With the exception of polygons, your `_fields` attribute simply adds `center`, `meta`, `visual`.  Can you simply use `_repr_params` and add [`center`, `meta`, `visual`] in the `copy` method?

Polygons would require a workaround, but I think the polygon API should be changed (but not in this release!) so that its "center" can be moved around by changing one parameter instead of all the vertices.  IIRC we discussed that at one time.

cdeil:
@larrybradley - Agreed that both `_fields` and `_repr_params` is bad. But I would prefer a tuple of fields, where for each field one says if it should be `repr=True` or `repr=False` as in https://docs.python.org/3/library/dataclasses.html . Not sure if we can achieve that for now.

Is it really useful to exclude `center` from `_repr_params`? Isn't this often the most important info about a region?

Another cosmetic change I'd suggest is to move these tuples to class-level, out of `__init__`. We don't need to access them from the class for now, but again I'd suggest to follow the design in https://docs.python.org/3/library/dataclasses.html a bit.

cdeil:
@larrybradley - I followed your suggestion in 19edcfc - for now we just have `_repr_params` as was already the case. I've written the copy in the base class so that it also works for polygons.

> Polygons would require a workaround, but I think the polygon API should be changed (but not in this release!) so that its "center" can be moved around by changing one parameter instead of all the vertices. IIRC we discussed that at one time.

You proposed it in #57. It wasn't clear to me why it's useful. Is it a performance issue, to have an efficient "shift" of a polygon to a new position, in case there are many vertices and you don't want to make another array?
I'm not really opposed to add a center or position attribute for polygons, but given that it does add some extra complexity (and probably isn't available for sky polygons?) - there should be a clear motivation for it.

cdeil:
Added rotate for circle, ellipse, rectangle and polygon. Seems to work fine, based on the test added.

Ready for final review.

larrybradley:
@cdeil I think the `_repr_params` didn't include the `center` property because it's assumed that most regions have a center, so the center was added once to `_repr_params` in the base class, i.e. `_repr_params` only included the parameters defining the shape, which differ from region to region.

I think it's a good idea to move the params tuple to the class level.  I'm not familiar with the `@dataclass` decorator, but the examples seem to all use type annotations, which I'm not a big fan of.

larrybradley:
@cdeil For the polygon region, I'm thinking of the use case of aperture photometry.  In that case, one has an `(x, y)` position/centroid of a source on which they want to perform photometry within a particular aperture.  Therefore, it would be nice for polygon regions to have the same API as the other regions that use `center`, e.g.:

```python
xypos = PixCoord(10, 20)
reg1 = CirclePixelRegion(xypos, radius=3)
reg2 = PolygonPixelRegion(xypos, vertices=vertices)
aperture_photometry(data, (reg1, reg2))
```

where the polygon `vertices` are defined *relative* to (0, 0).  In other words `vertices` defines the *shape* of the polygon and `position` defines its position.  That API mirrors all the other regions that separate the center parameter from the shape parameters, except we can't call it `center` for polygons because depending on how `vertices` are defined, it may not be the true center (although for aperture photometry it should be the true center).

```python
xypos2 = PixCoord(30, 40)
 # no need to recompute polygon vertices for new position
reg3 = PolygonPixelRegion(xypos2, vertices=vertices)
```

PR #279, opened 2019-06-16
Simplify annulus regions?

cdeil:
While working on the region copy methods in #269 I noticed that all "annulus" regions (see [here](https://astropy-regions.readthedocs.io/en/latest/shapes.html)) are implemented in a complex way and I was wondering if we should change the implementations.

E.g. [RectangleAnnulusPixelRegion](https://astropy-regions.readthedocs.io/en/v0.3/_modules/regions/shapes/annulus.html#RectangleAnnulusPixelRegion) creates `region1` and `region2` in `__init__` and then `AsymmetricAnnulusPixelRegion` is involved which has class-level variables like
```
inner_width = AnnulusInnerScalarLength('width')
```
which is a descriptor which reaches back via `getattr(instance, 'region1')` to find the number:
https://github.com/astropy/regions/blob/d1a6dd34def9442133065041974b2e33cafaf1cf/regions/core/attributes.py#L119-L149

It looks like this "annulus region = (outer region) xor (inner region)" suggestion came from
https://github.com/astropy/regions/pull/28#issuecomment-223495505 and then was introduced in #128 and later got more complicated when descriptors for validation were introduced and e.g. `AnnulusInnerScalarLength` was added.

I feel like what we have now is very complicated for no good reason, and we should change to something simple similar to [photutils.RectangularAnnulus](https://photutils.readthedocs.io/en/stable/api/photutils.RectangularAnnulus.html) where the numbers are simply stored in `__init__`. This would allow us to delete complex code like the `AnnulusInnerScalarLength` class completely. `contains` or `as_artist` could still use `outer - inner` implementation e.g. to avoid code duplication.

@keflavich @larrybradley - Thoughts on how to best implement those regions?

### DIFF
--- a/regions/shapes/annulus.py
+++ b/regions/shapes/annulus.py
@@ -16,7 +16,7 @@ from regions.core.attributes import ScalarPix, ScalarLength
 from regions.core.attributes import ScalarSky
 from .._utils.wcs_helpers import skycoord_to_pixel_scale_angle
 from ..core import PixelRegion, SkyRegion, PixCoord
-from ..shapes.circle import CirclePixelRegion, CircleSkyRegion
+from ..shapes.circle import CirclePixelRegion
 from ..shapes.ellipse import EllipsePixelRegion, EllipseSkyRegion
 from ..shapes.rectangle import RectanglePixelRegion, RectangleSkyRegion
 
@@ -174,17 +174,9 @@ class CircleAnnulusSkyRegion(SkyRegion):
 
 @six.add_metaclass(abc.ABCMeta)
 class AsymmetricAnnulusPixelRegion(AnnulusPixelRegion):
-    """
-    An asymmetrical annulus in pixel coordinates.
+    """Helper class for asymmetric annuli sky regions.
 
-    region1 : `~regions.PixelRegion` object
-        The inner asymmetric region
-    region2 : `~regions.PixelRegion` object
-        The outer asymmetric region
-    meta : `~regions.RegionMeta` object, optional
-        A dictionary which stores the meta attributes of this region.
-    visual : `~regions.RegionVisual` object, optional
-        A dictionary which stores the visual meta attributes of this region.
+    Used for ellipse and rectangle annuli below.
     """
 
     center = ScalarPix("center")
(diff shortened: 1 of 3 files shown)


cdeil:
I've refactored the annulus classes. 200 lines of code in `regions/core/attributes.py` are gone, all annulus related code is now in `regions/shapes/annulus.py`. The code there is still pretty complex (a class hierarchy), but I think this is a clear improvement - complex code in one file instead of distributed across two files that are far apart.

@keflavich or @sushobhana - could you please review this?

I'd like to get this in soon, and then continue and finish up the "copy" and "rotate" implementations for all regions, including these annulus regions.

cdeil:
The `AsymmetricAnnulusPixelRegion` is a base class for the ellipse and rectangle annulus regions.

Because they have 100% the same data members, and currently the implementation of `contains`, `as_artist`, ... is compound region based, they don't contain any code except for one method like this:
```
    def to_pixel(self, wcs):
        return EllipseAnnulusPixelRegion(
            *self.to_pixel_args(wcs), meta=self.meta, visual=self.visual
        )
```
which we could also move to the base class if we add a `_pix_class = EllipseAnnulusPixelRegion` class-level attribute. Should I do that?

I'm not happy with this deep inheritance hierarchy for code re-use, but also I don't have a simple suggestion at the moment how to avoid it without introducing code duplication.

Maybe merge this for now, and then hopefully in the future we will find a way to simplify further?

cdeil:
This is what we have now (almost the same as before, except that now the special descriptors in the other file are gone):

Screenshot 2019-06-16 at 14 37 15
Screenshot 2019-06-16 at 14 36 21

regions/shapes/annulus.py
```diff
@@ -197,7 +257,7 @@ def to_sky_args(self, wcs):
 
 
 @six.add_metaclass(abc.ABCMeta)
-class AsymmetricAnnulusSkyRegion(CompoundSkyRegion):
+class AsymmetricAnnulusSkyRegion(SkyRegion):
     """
     A rectangular annulus in `~astropy.coordinates.SkyCoord` coordinates.
```
keflavich: @cdeil I think it was this that threw me off; I didn't see the `RectangleAnnulus*` regions below that exist, which are totally fine.  This line should probably be
```suggestion
    An asymmetric annulus in `~astropy.coordinates.SkyCoord` coordinates; this is the base class for such region types (e.g., Rectangular and Elliptical).
```
or similar

keflavich: (again, I realize that's beyond what's in this PR... this does not necessarily need to be changed here, it just caught my eye when I was skimming)

larrybradley:
Late to the party, but thanks @cdeil.  FWIW, I also don't like the name `AsymmetricAnnulusPixelRegion`.  Why "asymmetric"?

PR #285, opened 2019-06-17
Implement rotate method for more regions

cdeil:
In this PR I finish implementing #265 by adding a rotate for all pixel regions.

The only region it's missing for is compound, I'll add that in this PR now.

@keflavich @astrofrog or @larrybradley - How do you feel about adding `PixelRegion.rotate`, i.e. something to the base class?

It could be an `abc.abstractmethod`, forcing all pixel regions to implement this. For us in Gammapy where we call `rotate` on user-provided regions, that would be a good thing. I would prefer that. But it would also mean that everyone subclassing `PixelRegion` has to implement it, or put this if they don't want to:
```
def rotate(center, angle):
    raise NotImplementedError
```

I could also put a generic implementation, that works for most built-in regions, allowing us to remove lines, i.e. put it as `PixelRegion.rotate` as a non-abstract method:
```
def rotate(center, angle):
        changes = {}
        changes["center"] = self.center.rotate(center, angle)
        if hasattr(self, "angle"):
            changes["angle"] = self.angle + angle
        # could add more here: "vertices", "start", "end"
        # could also scan `self._repr_params for PixCoord and transform those
        return self.copy(**changes)
```
But I think this is bad style, to rely on region specific attributes in the base class, and to silently return the wrong thing if some pixel region is different?

### DIFF
--- a/regions/core/compound.py
+++ b/regions/core/compound.py
@@ -165,6 +165,27 @@ class CompoundPixelRegion(PixelRegion):
     def area(self):
         raise NotImplementedError
 
+    def rotate(self, center, angle):
+        """Make a rotated region.
+
+        Rotates counter-clockwise for positive ``angle``.
+
+        Parameters
+        ----------
+        center : `PixCoord`
+            Rotation center point
+        angle : `~astropy.coordinates.Angle`
+            Rotation angle
+
+        Returns
+        -------
+        region : `CompoundPixelRegion`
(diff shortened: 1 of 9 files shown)


cdeil:
I'll merge this now as-is, so that we can try it in Gammapy. Rotate now works for all pixel regions, and there are tests for all of them. I did not add `PixRegion.rotate` on the base class.

PR #286, opened 2019-06-17
Change _repr_params to _fields, including center

cdeil:
This PR does some cleanup, following the discussion in 
 https://github.com/astropy/regions/pull/281#issuecomment-502833611

* Renames `_repr_params` to `_fields`, since it's now not just used for repr, but also copy
* Put `_fields` at the class level (before was `__init__`), right above the fields that are declared
* Add `center` to `_fields`

The last point might be a bit controversial, so here's my argument: explicit is better than implicit, and not all regions have a center (e.g. line and point and text and polygon), so before there was special code in the base class to treat those different cases, which I could now remove.

One more small change: the repr changed so that now `center=` appears. I think this is good, a repr should be explicit, it should show the attribute names. It's also more consitent now with `__str__` which already showed "center:" before.

### DIFF
--- a/regions/core/core.py
+++ b/regions/core/core.py
@@ -69,7 +69,7 @@ class Region(object):
 
     def copy(self, **changes):
         """Make an independent (deep) copy."""
-        fields = list(self._fields) + ["meta", "visual"]
+        fields = list(self._params) + ["meta", "visual"]
 
         for field in fields:
             if field not in changes:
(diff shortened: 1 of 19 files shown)


regions/core/core.py
```diff
@@ -98,11 +89,9 @@ def __repr__(self):
 
     def __str__(self):
         cls_info = [('Region', self.__class__.__name__)]
-        if hasattr(self, 'center'):
-            cls_info.append(('center', self.center))
-        if self._repr_params is not None:
+        if self._fields is not None:
```
larrybradley: I have a slight preference for `_params` over `_fields`, but no big deal.

cdeil: Changed to `_params`.

cdeil:
Doctests fixed (missed those tests since they apparently don't run when using `pytest` instead of `python setup.py pytest`).

CI should pass now. Ready for final review.

PR #293, opened 2019-07-03
Implement PolygonPixelRegion.area

cdeil:
This PR implements `PolygonPixelRegion.area` using basically https://stackoverflow.com/a/53864271/498873

#57 is a related with open question, but maybe we can put this in for now and then refactor depending on outcome of discussion there?


cdeil:
Merging this in now. I don't think it's controversial. We stil have #57 open to discuss polygons.

PR #304, opened 2019-10-21
Update to Python 3.6+ and Numpy 1.16+

cdeil:
This PR updates `astropy-regions` to Python 3.6+ and Numpy 1.16+.

To be in line with the Astropy core package (see [here](https://github.com/astropy/astropy/blob/bf4e4de4a6d1ceb1db10f5850a90cce476348903/setup.cfg#L28-L30).

x-ref #241 and #291

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,7 +62,7 @@ matrix:
 
         # Try older numpy versions
         - env: PYTHON_VERSION=3.6
-               NUMPY_VERSION=1.11
+               NUMPY_VERSION=1.16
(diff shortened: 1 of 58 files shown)


cdeil:
@larrybradley or @astrofrog - Please review.

cdeil:
I'll merge this in now, because I want to continue working on the `regions` package this week.

@larrybradley or @astrofrog or @keflavich or anyone - If you think we need Python 3.5 support or older Numpy version support than what Astropy core does, please shout and I'll revert some changes to make that work.

Issue #276, opened 2019-06-11
Add spherical circles and polygons?


cdeil:
This is a reminder issue that we might want to bring back spherical circles and polygons in the regions package.

We had at least spherical circle functionality at some point, but for consistency with DS9 and to avoid confusion of having one class represent a planar and a spherical sky circle sometimes, it was removed in #132 .

Now in #275 I'm removing the last trace of that functionality, an unused and untested helper function. But really, I think we probably should bring that functionality back (with a minor change, not call `numpy.matrix`, but instead use arrays and `numpy.dot` or `numpy.matmul`). I think it was discussed before, and we thought a separate class like `CircleSphericalSkyRegion` in addition to [CircleSkyRegion](https://astropy-regions.readthedocs.io/en/latest/api/regions.CircleSkyRegion.html) would be the way to go?
@astrofrog - Do you remember?

gpdf:
See #303; it would be helpful both in building client-side interfaces to IVOA services, but especially in implementing IVOA protocols in Python, if these spherical geometry region types existed, as they map directly to concepts in the IVOA interfaces.

cdeil:
The question how to represent and handle "true sky" regions that work without a WCS has come up many times, see e.g. #13, #295, #303, #76, #221

Ideas how to handle this included a flag (e.g. `mode={"planar", "spherical"}`) either on the region class, or on the methods that require this choice (namely contains and plotting), or separate classes.

I don't think there's a perfect solution, because the most-used serialisation format (DS9) doesn't support the distinction, and there is no great default behaviour. E.g. I think most users would expect [CircleSkyRegion](https://astropy-regions.readthedocs.io/en/latest/api/regions.CircleSkyRegion.html) to be spherical, but it's planar. Making it spherical would mean that one has to go to polygon on `sky_circle.to_pixel(wcs)` with some numerical precision (e.g. 100 points) and that wouldn't round-trip or be consistent with DS9 behaviour. So for those reasons we so far chose to remove the "spherical" variants completely and just do "planar" behaviour.

My suggestion would be to start by adding new classes for the spherical cases:
- `SphericalCircleSkyRegion`
- `SphericalPolygonSkyRegion`
- `SphericalRangeSkyRegion` (see #303)

It's contains would look like this, and it could or couldn't sub-class `CircleSkyRegion`, and e.g. `to_pixel` or `plot` would polygonise:
```
class SphericalCircleSkyRegion(CircleSkyRegion):
    """Spherical circle sky region."""

[...]

PR #306, opened 2019-10-23
Add spherical circle, range, polygon

cdeil:
I'd like to add spherical circle, range and polygon functionality to `astropy-regions.

This PR follows the thought in https://github.com/astropy/regions/issues/276#issuecomment-545361330 to add those as extra classes.

I'm not sure if that is the way to go, thus the PR in the early draft state.

For `RangeSphericalRegion` specifically I've left a few questions as `# TODO` - please make suggestions.

@larrybradley @astrofrog @eteq @gpdf - Thoughts?

### DIFF
--- a/regions/shapes/circle.py
+++ b/regions/shapes/circle.py
@@ -11,7 +11,7 @@ from .._geometry import circular_overlap_grid
 from ..core.attributes import (ScalarSky, ScalarPix, QuantityLength,
                                ScalarLength, RegionVisual, RegionMeta)
 
-__all__ = ['CirclePixelRegion', 'CircleSkyRegion']
+__all__ = ['CirclePixelRegion', 'CircleSkyRegion', 'CircleSphericalRegion']
 
 
 class CirclePixelRegion(PixelRegion):
@@ -195,3 +195,15 @@ class CircleSkyRegion(SkyRegion):
         center, scale, _ = skycoord_to_pixel_scale_angle(self.center, wcs)
         radius = self.radius.to('deg').value * scale
         return CirclePixelRegion(center, radius, self.meta, self.visual)
+
+
+class CircleSphericalRegion(CircleSkyRegion):
+    """Spherical circle sky region.
+
+    TODO: document
+    """
+
+    def contains(self, skycoord, wcs=None):
+        """Defined by spherical distance."""
+        separation = self.center.separation(skycoord)
+        return separation < self.radius
--- a/regions/shapes/range.py
+++ b/regions/shapes/range.py
@@ -1,0 +1,72 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+from astropy.coordinates import Angle, SkyCoord
+from regions.core.attributes import QuantityLength
+from ..core import SkyRegion, PixelRegion, PixCoord
+
+__all__ = ["RangePixelRegion", "RangeSphericalRegion"]
+
+inf = Angle(float("inf"), "deg")
+
+
+# TODO: needed?
+class RangePixelRegion(PixelRegion):
+    """
+    A range in pixel coordinates.
+
+    See `RangeSphericalRegion`.
+    """
+
+
+class RangeSphericalRegion(SkyRegion):
+    """
+    A range in longitude and latitude.
+
+    Use case: http://www.ivoa.net/documents/SIA/20151223/REC-SIA-2.0-20151223.html#toc12
+
+    Parameters
+    ----------
+    lon_min, lon_max : `~astropy.coordinates.Angle`
+        Range in longitude
+    lat_min, lat_max : `~astropy.coordinates.Angle`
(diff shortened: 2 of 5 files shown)


regions/shapes/range.py
```diff
@@ -0,0 +1,72 @@
...
+from astropy.coordinates import Angle, SkyCoord
+from regions.core.attributes import QuantityLength
+from ..core import SkyRegion, PixelRegion, PixCoord
+
+__all__ = ["RangePixelRegion", "RangeSphericalRegion"]
+
+inf = Angle(float("inf"), "deg")
+
+
+# TODO: needed?
```
gpdf: I don't think this is needed.  Isn't it covered exactly by the existing planar rectangle as long as you choose the right parameterization?  It might be thought of as a "convenience function".  Maybe an additional constructor / factory function for the planar rectangle?

regions/shapes/range.py
```diff
@@ -0,0 +1,72 @@
...
+        Range in latitude
+    frame : `~astropy.coordinates.BaseCoordinateFrame`
+        Sky frame
+    meta : `~regions.RegionMeta` object, optional
+        A dictionary which stores the meta attributes of this region.
+    visual : `~regions.RegionVisual` object, optional
+        A dictionary which stores the visual meta attributes of this region.
+    """
+    # TODO: should we use these properties, or SkyCoord for min / max,
+    # or a center-based representation?
```
gpdf: Please don't use a center-based representation.  That way lies the madness of ADQL's ill-fated BOX.

A true range-based representation is fundamentally not relocatable in latitude - its shape is inherently different for different latitudes.

A geometrically natural center-based representation needs to be fully relocatable - it should have the same shape anywhere on the sphere as long as its size parameters stay the same.  (It's a coordinate-free geometrical concept.)  

But it's *so* likely to be confused with the coordinate-range region that people have a lot of trouble with it.  BOX is being proposed for withdrawal from ADQL because it caused so much trouble of this kind.

regions/shapes/range.py
```diff
@@ -0,0 +1,72 @@
...
+    lon_min = QuantityLength('lon_min')
+    lon_max = QuantityLength('lon_max')
+    lat_min = QuantityLength('lat_min')
+    lat_max = QuantityLength('lat_max')
+
+    def __init__(self, lon_min=-inf, lon_max=inf, lat_min=-inf, lat_max=inf, frame="icrs", meta=None, visual=None):
+        self.lon_min = lon_min
+        self.lon_max = lon_max
+        self.lat_min = lat_min
+        self.lat_max = lat_max
```
gpdf: Python supports IEEE floating point -Inf and +Inf, so it should be possible, with care, to map these ranges exactly onto their IVOA SIAv2/SODA interpretation.  It may require a bit of thinking to handle the edge case of a region that actually covers the entire sky.

regions/shapes/circle.py
```diff
@@ -195,3 +195,15 @@ def to_pixel(self, wcs):
         center, scale, _ = skycoord_to_pixel_scale_angle(self.center, wcs)
         radius = self.radius.to('deg').value * scale
         return CirclePixelRegion(center, radius, self.meta, self.visual)
+
+
+class CircleSphericalRegion(CircleSkyRegion):
+    """Spherical circle sky region.
+
+    TODO: document
```
gpdf: See my comment on #276 .  I'd be happy to help review and/or write.

cdeil: @gpdf - Thanks!
I'll add some docs in this PR, and then you can review.

regions/shapes/range.py
cdeil: I think there's a few small advantages to introducing`RangePixelRegion` (instead of using RectanglePixelRegion):

- Clear code, symmetry, being able to round-trip when doing `to_pixel` and `from_pixel` isntead of mixing over to the rectangle classes
- I'm not sure how the +- inf case would be handled. I don't think one can handle the `min=-inf, max=42` case, there's no center that will result in a rectangle extending that range?
- It's not a lot of work (30 lines of code?)

I'm still not sure if it's worth adding, but my feeling is that it might be the best way to go to add it?

@gpdf - Thoughts?

If no addition, what exactly whould be the alternative? 
Should `RangeSphericalRegion` not offer `to_pixel` at all, or should it go to `RectanglePixelRegion`?

regions/shapes/range.py
cdeil: OK, I'll add test cases.

regions/shapes/range.py
cdeil: OK, I'll leave the current representation then, as four angles and a separate frame attribute.

cdeil:
Unfortunately I don't have time for this any more, closing it now. See https://github.com/astropy/regions/issues/276#issuecomment-567005691.

PR #314, opened 2020-05-14
Fix for issue 312

keflavich:
Issue 312 showed that writing a region with no metadata resulted in an invalid CASA region file with a trailing comma.

 * [x] TODO: add a test


JonahDW:
Tested the code, and indeed no more trailing comma! However CASA doesn't recognize the coordinate system. I believe using 
"coord={}, ".format(coordsys_mapping['CRTF'][coordsys])

Should map the input coordinate system to one that CASA accepts and can use.

JonahDW:
I think that about does the trick! I think it would makes sense to rebase my PR against this and continue working there?

keflavich:
:+1: yes do that, but I need to fix the test here, I think

keflavich:
I broke the tests with this change because the tests were wrong, but I think it's doing a coordinate transformation now, which doesn't make a lot of sense...

keflavich:
oops, tests had FK4 in output, which isn't allowed in the spec.  We should really test our "correct" file against CASA...

regions/io/crtf/tests/test_crtf_language.py
```diff
@@ -5,11 +5,15 @@
 import astropy.version as astrov
 from astropy.utils.data import get_pkg_data_filename
 
+from astropy import coordinates, units as u
+
```
larrybradley, line 9: Extra blank line here.

PR #317, opened 2020-05-21
Make it possible to initialize and update regions based on interactive Matplotlib widgets

astrofrog:
This is a proof of concept of initializing and updating regions based on interactive Matplotlib widgets (which are now a think in ``matplotlib.widgets``). The idea is that regions could have a ``as_mpl_selector``  method that takes the axes and returns a Matplotlib selector. The selector is then active and the region is updated as the selector is moved around and adjusted (provided ``sync`` is set to ``True``, which is the default). Here's an example in action:

```python
import numpy as np
from regions import RectanglePixelRegion, PixCoord

data = np.random.random((16, 16))
mask = np.zeros_like(data)

plt.ion()

ax1 = plt.subplot(1, 2, 1)
ax1.imshow(data)
ax2 = plt.subplot(1, 2, 2)
im_mask = ax2.imshow(mask, vmin=0, vmax=1)


def update_mask(region):
    mask = region.to_mask(mode='subpixels', subpixels=10).to_image(data.shape)
    im_mask.set_data(mask)


region = RectanglePixelRegion(PixCoord(3, 5), 2, 3)
selector = region.as_mpl_selector(ax1, callback=update_mask, sync=True)

plt.show(block=True)
```
Peek 2020-05-21 13-41

The Matplotlib widgets included in Matplotlib don't (yet) have everything we need, in particular they don't have the ability to support rotation. However, I'm going to investigate whether the Matplotlib developers would be happy to have this additional functionality. Also moving regions around is a bit obnoxious as it requires clicking on the center of the region. I'm going to see if I can patch it to allow the whole region to be clicked and dragged. But in any case, we can either improve the widgets in Matplotlib or fork it here if they won't accept improvements upstream (which I doubt).

@keflavich @larrybradley - what do you think about this approach and API?

TODOs if we agree on the approach:

* [x] Use the region's visual properties for the selector
* [x] Changelog entry
* [x] Tests
* [ ] Expand approach to other region types
* [ ] Investigate how to deal with Sky regions and WCSAxes
* [ ] Improve Matplotlib selectors to support e.g. rotation (could potentially be done as part of another PR since we can just raise a NotImplementedError in the mean time)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,10 @@
   ``pyproject.toml`` to opt in to isolated builds as described in PEP 517/518.
   [#315]
 
+- Added a ``as_mpl_selector`` method to the rectangular and ellipse
+  pixel-based regions. This method returns an interactive Matplotlib
+  selector widget. [#317]
+
 0.4 (2019-06-17)
 ================
 
--- a/regions/shapes/ellipse.py
+++ b/regions/shapes/ellipse.py
@@ -268,7 +268,11 @@ class EllipsePixelRegion(PixelRegion):
             def sync_callback(*args, **kwargs):
                 pass
 
-        self._mpl_selector = EllipseSelector(ax, sync_callback, interactive=True)
+        self._mpl_selector = EllipseSelector(ax, sync_callback, interactive=True,
+                                             rectprops={'edgecolor': self.visual.get('color', 'black'),
+                                                        'facecolor': 'none',
+                                                        'linewidth': self.visual.get('linewidth', 1),
+                                                        'linestyle': self.visual.get('linestyle', 'solid')})
         self._mpl_selector.extents = (self.center.x - self.width / 2,
                                       self.center.x + self.width / 2,
                                       self.center.y - self.height / 2,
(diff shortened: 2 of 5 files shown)


keflavich:
Sounds like a great approach.  I very much like the mpl upstream push too, though that is likely to slow adoption as mpl had historically been one of the slower packages to update.

larrybradley:
@astrofrog, that looks great.  How would ellipses be drawn?  If we add rotation, we'll also need a way to do that with the cursor (the usual API is to select and drag box corners).  For ellipses, I assume there would be something to select on the major (and minor) axes?

astrofrog:
@larrybradley - the ellipse selector is very similar (just added):

ellipse

For rotation, as in some other programs, I would envision pressing a modifier key and selecting any of the anchors and dragging it around. Or we could make one of the corners be a rotation one.

astrofrog:
I’m chasing down the remaining CI failures

astrofrog:
@keflavich @larrybradley - this is now ready for review!

keflavich:
lgtm!  awesome!  It deserves some documentation, but this is *great*!

PR #319, opened 2020-05-27
DO NOT MERGE: Test with astropy 4.1rc1

pllim:
Test with astropy 4.1rc1


pllim:
CircleCI seems unrelated. Hard to tell from the Travis CI tests due to lack of pytest-astropy-header but seems to have installed 4.1rc1 based on the raw logs.

PR #329, opened 2020-11-30
DOC: Replace astropy channel with conda-forge channel 

pllim:
Similar to astropy/astropy#11111


larrybradley:
Thanks, @pllim.  The `build_docs` failure is because `Shapely` cannot be installed with Python 3.9 (no linux/osx wheels yet).

PR #337, opened 2021-03-29
Update Cython version used in pyproject.toml

astrofrog:
I think this might fix some of the Windows CI issues

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
 requires = ["setuptools",
             "setuptools_scm",
             "wheel",
-            "cython==0.29.14",
+            "cython==0.29.22",
             "oldest-supported-numpy",
             "extension-helpers"]
 build-backend = 'setuptools.build_meta'


larrybradley:
The test failures look real.  `TestEllipsePixelRegion.test_as_mpl_selector` is expecting a center of 8.3, but getting 5.0.
https://dev.azure.com/astropy-project/regions/_build/results?buildId=1146&view=logs&j=7f33e5bd-7764-5d8a-ba2e-506e078b9c3f&t=57e5be12-0a85-50b9-0e83-1cf5152ef732&l=218

astrofrog:
I wonder if this is related to some changes in Matplotlib

astrofrog:
I can't reproduce the issue locally, I wonder whether there is some kind of race condition in the interactivity or if we somehow need to force everything to be drawn before checking values.

astrofrog:
This does fix the docs build which had the Cython issues though so going to merge this and we should investigate the selector test issues separately (these happen even without the Cython upgrade)

PR #388, opened 2021-07-23
Remove wheels from CI tests

larrybradley:
For the v0.5 release I ended up building the wheels using other tools.  Having the wheels in the CI tests here takes a long time and wastes CPU cycles.


astrofrog:
I do think it's nice to have the wheel building live in the same repo but we could try and set up a system more like in the core package where we only build on specific branches or tags (and we could also use GitHub actions instead of Azure). No need to do anything for now but once I have spare cycles I'd like to push for the same infrastructure to be used across the astropy org.

PR #396, opened 2021-08-06
Add ability to add and subtract PixCoord objects

larrybradley:
Examples:
```python
>>> p1 = PixCoord([3, 3], [4, 4])
>>> p2 = PixCoord([5, 7], [6, 1])
>>> p1 + p2
PixCoord(x=[ 8 10], y=[10  5])

>>> p1 - p2
PixCoord(x=[-2 -4], y=[-2  3])

>>> PixCoord([3, 3], [4, 4]) + PixCoord(10, 10)
PixCoord(x=[13 13], y=[14 14])
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,8 @@ New Features
 
 - Added the DS9 'boxcircle' point symbol. [#387]
 
+- Added the ability to add and subtract ``PixCoord`` objects. [#396]
+
 Bug Fixes
 ---------
 
--- a/regions/core/pixcoord.py
+++ b/regions/core/pixcoord.py
@@ -121,16 +121,25 @@ class PixCoord:
         y = self.y[key]
         return PixCoord(x=x, y=y)
 
+    def __add__(self, other):
+        if not isinstance(other, self.__class__):
+            raise TypeError('Can add only to another PixCoord')
+        return self.__class__(self.x + other.x, self.y + other.y)
+
+    def __sub__(self, other):
+        if not isinstance(other, self.__class__):
+            raise TypeError('Can subtract only from another PixCoord')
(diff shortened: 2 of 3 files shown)


keflavich:
I like this, but it should be documented.  Is the plan to also allow subtraction of `PixCoord` objects from any/all pixel shapes?  Maybe that already exists...

larrybradley:
You mean subtract a `PixCoord` from a `PixelRegion` object to shift it?  I wasn't thinking about that (and don't think I'd be in favor of it).

larrybradley:
I added a note to the docstring in 00e31ab36903f9cd43b9b7affed591eed140c17d.

keflavich:
Why are you opposed to using subtraction to shift?  Isn't that the point of this PR?  Or is there another use case here I'm not thinking of?

larrybradley:
I needed this for https://github.com/astropy/regions/pull/397, where a `PixCoord` is added to a `PixCoord` (https://github.com/astropy/regions/blob/main/regions/shapes/polygon.py#L72).

`reg = CirclePixelRegion(...) + PixCoord(...)` seems a bit awkward to me (adding two different object types).  Instead, I would propose a `PixelRegion.shift(...)` method, e.g., `CirclePixelRegion(...).shift(PixCoord(...))`, analogous to the existing `PixelRegion.rotate(...)` method.

keflavich:
Adding/subtracting different object types doesn't seem bad to me, but I'm not going to push it.

I think we'd want both `XPixelRegion.shift(dx, dy)` and ` XPixelRegion.separation(PixelRegion)`... though the latter only makes sense for `PixelRegion` and others that have a defined center.  I do think that measuring separations between regions could be useful.

Anyway, this is off-topic, I don't have any objections to the changes you've made now that I understand the motivation.

PR #398, opened 2021-08-06
Add RegularPolygonPixelRegion class

larrybradley:
This PR adds a `RegularPolygonPixelRegion` class defined by a center, number of vertices (sides), radius, and optional angle.

Examples:

regions-RegularPolygonPixelRegion-1

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,8 @@ New Features
 - Added an ``origin`` keyword to ``PolygonPixelRegion`` to allow
   specifying the vertices relative to an origin pixel. [#397]
 
+- Added a ``RegularPolygonPixelRegion`` class. [#398]
+
 Bug Fixes
 ---------
 
--- a/regions/io/core.py
+++ b/regions/io/core.py
@@ -13,7 +13,8 @@ import numpy as np
 from ..shapes import (CirclePixelRegion, CircleSkyRegion,
                       EllipsePixelRegion, EllipseSkyRegion,
                       RectanglePixelRegion, RectangleSkyRegion,
-                      PolygonPixelRegion, PolygonSkyRegion,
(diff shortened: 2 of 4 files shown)


keflavich:
Does this region type have any round-trippable serializations?  I don't think ds9 or CRTF support it, but perhaps the VO serialization format does?  If not, do we have a list of regions that can('t) be round-trip-serialized?  We should.  This is probably something to raise as a separate issue.

larrybradley:
Serialization for this object doesn't work yet, but `RegularPolygonPixelRegion` is subclassed from `PolygonPixelRegion` so I can serialize/write it to ds9 using the polygon shape.  Reading/writing it from ds9 would then return a `PolygonPixelRegion` instead of a `RegularPolygonPixelRegion` -- so not exactly round-trippable to give the same *object*, but the defined region would be the same.

keflavich:
Yes, that's exactly what I meant - a user can serialize it just fine, but they will not get back a similarly-manipulatable object if they read it from ds9/carta.

larrybradley:
We could always define our own serialization format (perhaps using ASDF?), but that's not a priority.

keflavich:
I'm not suggesting [that](https://xkcd.com/927/)! 😱   I'm just saying we should note it!

larrybradley:
We *could* also add a method converts PolygonPixelRegion to RegularPolygonPixelRegion (if it's a regular polygon).

larrybradley:
I've fixed this serialization and added a note in the docstring about the round-trip serialization:

Screen Shot 2021-08-05 at 10 59 42 PM

keflavich:
For the regular polygon 'recognizer', we'd be subject to precision limitations in the serialization, which would mean probably introducing some threshold for 'regular'.  That's fine, just awkward.

Issue #391, opened 2021-07-30
Outstanding work on interactive regions functionality


dhomeier:
Opening this just to keep track of the task list outlined and partly completed in #317:

> The Matplotlib widgets included in Matplotlib don't (yet) have everything we need, in particular they don't have the ability to support rotation. However, I'm going to investigate whether the Matplotlib developers would be happy to have this additional functionality. Also moving regions around is a bit obnoxious as it requires clicking on the center of the region. I'm going to see if I can patch it to allow the whole region to be clicked and dragged. But in any case, we can either improve the widgets in Matplotlib or fork it here if they won't accept improvements upstream (which I doubt).
>
> @keflavich @larrybradley - what do you think about this approach and API?
>
> TODOs if we agree on the approach:

- [x]   Use the region's visual properties for the selector
- [x]  Changelog entry
- [x]  Tests
- [ ]  Expand approach to other region types
- [ ]  Investigate how to deal with Sky regions and WCSAxes
- [x]  Improve Matplotlib selectors to support e.g. rotation (could potentially be done as part of another PR since we can just raise a NotImplementedError in the mean time)

The last point is under development in #390.

Re. _allow the whole region to be clicked and dragged_ this is already implemented by holding `space` and seems to just work, but the matplotlib documentation feels a bit unclear about it. The modifier key is defined for `_SelectorWidget` in
https://github.com/matplotlib/matplotlib/blob/01fb7bb2d9a674a05f704fc50d36e0baa7d6079c/lib/matplotlib/widgets.py#L1798-L1799
but the docstring e.g. for `RectangleSelector` states
https://github.com/matplotlib/matplotlib/blob/01fb7bb2d9a674a05f704fc50d36e0baa7d6079c/lib/matplotlib/widgets.py#L2626-L2634
where possibly `' '` has simply been confused for `None`. Ping the matplotlib devs for clarification?

Additionally https://github.com/matplotlib/matplotlib/pull/19657 has added an option to always enable this behaviour; currently this can be activated by
```python
selector.drag_from_anywhere = True
```
so all that remains to be done there would be to let `region.as_mpl_selector` pass this kwarg to the Selector on creation.

keflavich:
[...]

PR #406, opened 2021-12-08
Add interactive matplotlib selector to `PolygonPixelRegion`

dhomeier:
As next part of the work on interactive regions #391 this PR adds an `as_mpl_selector` widget to the `PolygonPixelRegion` class to allow moving and resizing through the `draw_box` functionality added to `PolygonSelector` in matplotlib/matplotlib#21830, as well as reshaping by direct manipulation of the vertices. The `draw_box` handles are working analogously to the selector methods for Rectangle and Ellipse PixelRegions.
Rotation is not yet included, but can be added once the mpl PR or a follow-up supports this. A `centroid` method is already included to provide a robust and rotation-invariant reference point.
The connected `PolygonSelector` is instantiated in a rather pedestrian manner, using several private methods, which should perhaps be replaced by a public method on the matplotlib side.

### DIFF
--- a/regions/shapes/polygon.py
+++ b/regions/shapes/polygon.py
@@ -7,8 +7,7 @@ import astropy.units as u
 from astropy.wcs.utils import pixel_to_skycoord, skycoord_to_pixel
 import numpy as np
 
-from ..core.attributes import (OneDPix, OneDSky, ScalarPix, ScalarLength,
-                               QuantityLength)
+from ..core.attributes import OneDPix, OneDSky, ScalarPix, ScalarLength, QuantityLength
 from ..core.bounding_box import BoundingBox
 from ..core.core import PixelRegion, SkyRegion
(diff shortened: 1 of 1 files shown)


dhomeier:
The last commit adds a method to set the rotation angle on a given Polygon instance as `polygon.rotation = 30 * u.deg`, and update the selector from it, if present (no test for the latter as the mpl PR is not merged) – as kind of a template for how rotation from the selector might eventually be implemented.

pllim:
If https://github.com/astropy/regions/pull/500 is merged first, the doc needs to be updated here too.

dhomeier:
Status from December 2024 rebased to current main; to proceed further polygon rotation needs to be added to matplotlib, which should probably be done after the corresponding functionality for rectangles and ellipses in matplotlib/matplotlib#26833 has been merged.

PR #408, opened 2021-12-14
Fix for Issue 407: CRTF region parsing

keflavich:
CRTFs appear to interpret :'s as hour-separating, .'s as degree-separating.

### DIFF
--- a/regions/io/crtf/read.py
+++ b/regions/io/crtf/read.py
@@ -482,10 +482,20 @@ class _CRTFCoordinateParser:
             return u.Quantity(string_rep[:-3], u.dimensionless_unscaled)
         if 'h' in string_rep or 'rad' in string_rep:
             return Angle(string_rep)
+
+        # CRTF format apparently implicitly defines
+        # 18:20:30.12
+        # strings as hour units, and
+        # 10.11.54.69
+        # as degrees.
+        # I can't find this definition in documentation anywhere, though.
         if len(string_rep.split('.')) >= 3:
             string_rep = string_rep.replace('.', ':', 2)
+            unit = u.deg
+        elif string_rep.count(":") == 2:
+            unit = u.hour
 
-        return Angle(string_rep, u.deg)
+        return Angle(string_rep, unit)
 
     @staticmethod
     def parse_angular_length_quantity(string_rep):
--- a/regions/io/crtf/read.py
+++ b/regions/io/crtf/read.py
@@ -489,9 +489,9 @@ class _CRTFCoordinateParser:
         # 10.11.54.69
         # as degrees.
         # I can't find this definition in documentation anywhere, though.
+        unit = u.deg
         if len(string_rep.split('.')) >= 3:
             string_rep = string_rep.replace('.', ':', 2)
-            unit = u.deg
         elif string_rep.count(":") == 2:
             unit = u.hour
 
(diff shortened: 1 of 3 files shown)


regions/io/crtf/read.py
```diff
@@ -482,10 +482,20 @@ def parse_coordinate(string_rep):
...
+        # 18:20:30.12
+        # strings as hour units, and
+        # 10.11.54.69
+        # as degrees.
+        # I can't find this definition in documentation anywhere, though.
         if len(string_rep.split('.')) >= 3:
             string_rep = string_rep.replace('.', ':', 2)
+            unit = u.deg
+        elif string_rep.count(":") == 2:
+            unit = u.hour
```
dhomeier, line 496:
```suggestion
        unit = u.deg
        if len(string_rep.split('.')) >= 3:
            string_rep = string_rep.replace('.', ':', 2)
        elif string_rep.count(":") == 2:
            unit = u.hour
```
Should there be a default `unit`, e.g. shall 10.119 be parsed as `10.119 * u.deg`?
Thought that was causing the test to fail, but actually it is just not raising the right exception now, expecting input like '18h12m24s' to raise `CRTFRegionParserError`, getting `undefined variable 'unit'` instead.

dhomeier (changes requested):
Looks OK apart from either fixing the failing test and/or setting a default unit.
Defaulting to `hms` for RA makes sense anyway, but it's a bit strange the format is not clearly defined. I think right now an input like
> 18:20:30.12 4:11:54.69
would parse declination as `hms` as well; should it raise an invalid format error instead or default to `dms` for declination?

regions/io/crtf/tests/test_crtf.py
```diff
@@ -204,3 +204,11 @@ def test_angle_serialization():
...
                 '20.000002deg], 0.000278deg]\n')
     assert regstr == expected
+
+
+def test_read_sexagesimal_regression407():
+    filename = 'data/crtf_carta_sexagesimal.crtf'
+    filename = get_pkg_data_filename(filename)
+    regs = Regions.read(filename, errors='warn', format='crtf')
+
+    assert regs[0].center.ra == 281.93342096*u.deg
```
dhomeier, line 214:
```suggestion
    assert_quantity_allclose(regs[0].center.ra, 281.93342096*u.deg, rtol=1e-9)
```
Looked a bit suspicious with the float comparison, but that was a nasty one lurking just past the default printed digits! Will need
`from astropy.tests.helper import assert_quantity_allclose`
at the top (though `np.allclose` should work now with quantities for all supported numpy versions as well).

dmehring:
In the hope that this will help clarify CRTF.

Angular measures of the form 05h30m22s represent  angular measures using time syntax, in hours, minutes, seconds eg, RA. Angular measures of the form 5d30m22s represent angles in degrees, arcmin, arcsec. The leading "h" or "d" determines the type.

From a previous comment:

"18:20:30.12 4:11:54.69"

In CRTF format, if the second measure is intended to represent a declination in the usual degrees,arcmin,arcsec, it should be written as 4.11.54.89. The first form (with : delimiters) represents angular measures using time syntax, the second represents angles in degrees.arcmin.arcsec.fraction_of_arcsec

larrybradley (approved):
Thanks, @keflavich!

PR #412, opened 2022-01-06
Fix codestyle errors

larrybradley:
~~This test requires python 3.6.  Also, I did some digging and this test was *never* running because `casatools` was not installed properly.  In fact, it cannot even pass because the `me` variable is never defined.~~

I've kept the test, adding `from casatools import measures as me`.

### DIFF
--- a/regions/io/crtf/tests/test_casa_mask.py
+++ b/regions/io/crtf/tests/test_casa_mask.py
@@ -12,6 +12,7 @@ from ..write import _write_crtf
 
 try:
     from casatools import image, simulator
+    from casatools import measures as me
     from casatasks import tclean
     HAS_CASATOOLS = True
 except ImportError:
--- a/regions/io/crtf/tests/test_crtf.py
+++ b/regions/io/crtf/tests/test_crtf.py
@@ -198,7 +198,7 @@ def test_angle_serialization():
     Regression test for issue #223 to ensure Angle arcsec inputs are
     correctly converted to degrees.
     """
-    reg = Regions([CircleSkyRegion(SkyCoord(10,20, unit='deg'),
+    reg = Regions([CircleSkyRegion(SkyCoord(10, 20, unit='deg'),
                                    Angle(1, 'arcsec'))])
     regstr = reg.serialize(format='crtf')
     expected = ('#CRTFv0\nglobal coord=J2000\ncircle[[10.000009deg, '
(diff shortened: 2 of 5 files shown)


larrybradley:
CI runs showing the test was always skipped:

https://dev.azure.com/astropy-project/regions/_build/results?buildId=1926&view=logs&j=0a22f5d8-9f3e-5735-4672-5a55fb1a2abf&t=7886e472-715d-500e-d95f-9ccab8aa9bc3&l=119

Error when you don't skip the test:

```
=================================== FAILURES ===================================
______________________________ test_casa_masking _______________________________

    @pytest.mark.skipif('not HAS_CASATOOLS')
    def test_casa_masking():
        with tempfile.TemporaryDirectory() as tmpdir:
            # SIMULATE SOME DATA SET

            # Define antennas
            diam = [25, 25, 25, 25, 25]
            xx = [50, 100, 150, 200, 250]
            yy = [2, -5, -20, -50, -100]
            zz = [-0.5, -1.0, -1.5, -2.0, -2.5]

            sm = simulator()
            sm.open(tmpdir + '/SIM.ms')
            # do configuration
>           posvla = me.observatory('VLA')
E           NameError: name 'me' is not defined

crtf/tests/test_casa_mask.py:35: NameError
=========================== short test summary info ============================
FAILED crtf/tests/test_casa_mask.py::test_casa_masking - NameError: name 'me'...
=================== 1 failed, 63 passed, 3 xfailed in 27.50s ===================
```

keflavich:
I'd like to keep this test, but fix it.  Could you just mark it as "skip" instead?  Then I'll revisit it.

larrybradley:
Doesn't CASA require Python <= 3.6?

keflavich:
not any more!  It used to, though, which is why we had the test pegged that way.

larrybradley:
Currently, the `codestyle` tests are also failing because of this test (the linter is picking up that `me` is an undefined variable).  I can mark the test as "xfail", but I'll need to make the linter happy.

keflavich:
add `me = 'sad'` ?

keflavich:
(`me` is supposed to be the measurements package)

larrybradley:
CI is all green now, so merging.

There is a test (`TestRectanglePixelRegion.test_as_mpl_selector`) failing with dev dependencies (latest numpy and astropy).  That is an allowed failure, but we should investigate and fix it sooner than later.

PR #419, opened 2022-01-28
Add documentation section and example for using `as_mpl_selector` with callback function

dhomeier:
This adds a new subsection to the `shapes` docs, and an alternative method to the `masks` examples.
The latter is interactive by nature; have to live with a still for now. Also it cannot demonstrate a rotated ellipse until #390 is merged. And the Polygon version will only be available with #406.


keflavich (approved):
This is great, thanks!

larrybradley (approved):
Thanks, @dhomeier!

PR #421, opened 2022-02-02
Add the ability to compare Region objects

larrybradley:
This PR adds the ability to compare `Region` objects for equality (or inequality).  This is especially useful for testing.


larrybradley:
Test failure is unrelated (region selectors)

Issue #56, opened 2016-07-21
Defining regions without centers


larrybradley:
It would be nice to be able to define regions without specifying the center positions (or have a default center of e.g. (0,0)).  This issue came up when trying to define an object-oriented interface for local background subtraction for `photutils` PSF fitting, where the positions of the stars would be determined in a downstream function, e.g.:

```
from photutils import MMMBackground, CircularAnnulus, psf_photometry
aper = CircularAnnulus(r_in=15, r_out=30)       # position of stars to be determined later
bkg = MMMBackground( sigma=3.)
psf_phot_table = psf_photometry(data, ...., background_sub=bkg, background_aperture=aper)
```

Inside the `psf_photometry` function, the apertures positions could then be defined by setting the object `center` attribute.

cdeil:
@larrybradley - It looks like this is currently possible:
```python
In [1]: from regions import CircleSkyRegion, CirclePixelRegion, PixCoord

In [2]: from astropy.coordinates import Angle, SkyCoord

In [3]: CircleSkyRegion(center=None, radius=None)
Out[3]: 
CircleSkyRegion
Center: None
Radius: None

In [4]: CirclePixelRegion(center=None, radius=None)
Out[4]: 
CirclePixelRegion
Center: None
Radius: None
```

But I think that's only because no-one got around to implementing input validation yet.
So if this is important behaviour, we should establish it via a unit test that contains a comment that this is on purpose.

I'm putting this on the v0.2 milestone for now.

@larrybradley - Do you have time for this in the coming weeks?

larrybradley:
@cdeil Ah, great!  Thanks.

Sure I can add a unit test to ensure this capability.

larrybradley:
Should we make the `center` and `radius` default to `None` so that one can initialize a region with no arguments, e.g. `CircleSkyRegion()`?

cdeil:
Ah, I see, currently it's this:
```python
class CircleSkyRegion(SkyRegion):
    """
    A circle in sky coordinates.

    Parameters
    ----------
    center : `~astropy.coordinates.SkyCoord`
        Center position
    radius : `~astropy.units.Quantity`
        Radius in angular units
    """

    def __init__(self, center, radius, meta=None, visual=None):
        # TODO: test that center is a 0D SkyCoord
        self.center = center
        self.radius = radius
        self.meta = meta or {}
        self.visual = visual or {}
```
[...]

PR #430, opened 2022-02-16
Require radius, width, and height parameters to be strictly positive (> 0)

larrybradley:
Regions with zero or negative sizes (eventually) raise errors and are not useable.

This PR also updates the descriptor classes (and renames many of them, but they are not part of the public API).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -68,6 +68,9 @@ API Changes
 - Renamed the ``BoundingBox`` class to ``RegionBoundingBox``. The old
   name is deprecated. [#427]
 
+- A ``ValueError`` is raised if the radius, width, or height region
+  parameters are not strictly positive (> 0). [#430]
+
 
 0.5 (2021-07-20)
 ================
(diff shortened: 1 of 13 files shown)


keflavich (approved):

adonath:
Sorry for being late here...I just noticed that this change breaks some of our CI: https://github.com/gammapy/gammapy/runs/5649993418?check_suite_focus=true#step:10:5614

This is because in some cases we use `np.nan` as placeholder for the `width/height` argument if the information is not available. This is convenient because it allows to process everything in a uniform way, without having to catch errors. E.g., extracting an array of source extensions from a list of regions, of which some are `np.nan`, which can easily happen for non-uniform catalogs or loose data collections. So my main question here is whether not allowing `np.nan` either was an intended change here?

larrybradley:
I wasn't thinking specifically about `np.nan` inputs, but I wasn't expecting them either.  

I raised the topic of not setting shape parameters (`center`, specifically) almost 6 years ago back in https://github.com/astropy/regions/issues/56, but it was never implemented.  I think this needs discussion with the other devs.  I'll open a new issue.

adonath:
Thanks @larrybradley, #446 is indeed a very similar request.

PR #431, opened 2022-02-17
Add shape parameter descriptions to be used as attribute docs

larrybradley:
Currently the parameter attributes like `center`, `radius`, `width`, `height`, and `angle` have non-useful docstrings, which taken from the docstrings of the descriptor classes, e.g.:

docs

This PR adds a description keyword to the descriptor classes, which is used for the attribute docstrings.  This PR also simplifies the example code in the docstrings.


larrybradley:
CI failure is unrelated

PR #435, opened 2022-03-02
DOC: Clarify ellipse semimajor/minor axes

pllim:
This is to be more consistent with photutils. See https://photutils.readthedocs.io/en/stable/api/photutils.aperture.EllipticalAperture.html#photutils.aperture.EllipticalAperture

I had to code dive a little to find out, so we should add this to docstring.

### DIFF
--- a/regions/shapes/ellipse.py
+++ b/regions/shapes/ellipse.py
@@ -33,9 +33,9 @@ class EllipsePixelRegion(PixelRegion):
     center : `~regions.PixCoord`
         The position of the center of the ellipse.
     width : float
-        The width (semimajor axis) of the ellipse (before rotation) in pixels.
+        The width (twice the semimajor axis) of the ellipse (before rotation) in pixels.
     height : float
-        The height (semiminor axis) of the ellipse (before rotation) in pixels.
+        The height (twice the semiminor axis) of the ellipse (before rotation) in pixels.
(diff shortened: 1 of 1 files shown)


keflavich (approved):

larrybradley:
I understand the confusion here, but I think the names semimajor and semiminor were deliberately left off here.  I think it was @astrofrog (?) who initially created these region classes, based from the `photutils` aperture classes.  There are no checks anywhere to ensure that `width >= height`, which would be needed to assume `width = semimajor` and `height = semiminor`.  Admittedly that's how I mentally think of them, but there's no restriction of having `height` be the semimajor axis (which implies a rotation of 90 deg).

I think it does make more sense to call these elliptical parameters as semimajor and semiminor axes (as photutils does).  But that decision was made before I started working on this package.  My best guess is that perhaps this was done to make the API more consistent with rectangles (width/height)?  Perhaps @astrofrog remembers?

pllim:
Well, from the way you calculate stuff in the code, they correspond one-to-one with the way photutils defines semimajor/semiminor. 🤷

larrybradley:
Yep, I know.  All of the low-level Cython code from `photutils` (that @astrofrog originally wrote) was copied to `regions`.

larrybradley:
Closing since this a feature, not a bug.

PR #440, opened 2022-03-07
Update attribute descriptors to use __set_name__

larrybradley:
This PR updates the descriptors (used as region attribute validators) to use `__set_name__` (PEP-487).  It also adds attribute docs for the ellipse/rectangle pixel/sky regions, which were not being rendered by Sphinx.  It also improves the docs for several attributes.


larrybradley:
This CI failure needs to be fixed in `astropy` (https://github.com/astropy/astropy/issues/12929)

PR #462, opened 2022-05-24
Allow region meta or visual keywords to be a dict

### LINKED ISSUE #461: Initializing a region with visual={} results in an error when plotting
For a region defined like this:
```python
reg3 = regions.Regions([regions.CircleSkyRegion(center=SkyCoord('18:46:03.78', '-2:39:22.435', frame='icrs', unit=(u.h, u.deg)), radius=0.15*u.arcsec, visual={'color':'r'})])
```

the following traceback occurs:
```python
AttributeError                            Traceback (most recent call last)
/scratch/local/34157334/ipykernel_60949/2692360403.py in <module>
----> 7 reg3[0].to_pixel(cont[0].wcs).plot(ax=ax)

/blue/adamginsburg/adamginsburg/repos/regions/regions/core/core.py in plot(self, origin, ax, **kwargs)
    402             ax = plt.gca()
    403 
--> 404         artist = self.as_artist(origin=origin, **kwargs)
    405         ax.add_artist(artist)
    406 

/blue/adamginsburg/adamginsburg/repos/regions/regions/shapes/circle.py in as_artist(self, origin, **kwargs)
    152         xy = self.center.x - origin[0], self.center.y - origin[1]
    153         radius = self.radius
--> 154         mpl_kwargs = self.visual.define_mpl_kwargs(self._mpl_artist)
    155         mpl_kwargs.update(kwargs)
    156 

AttributeError: 'dict' object has no attribute 'define_mpl_kwargs'
```

The workaround is to specify the `visual` as a `RegionVisual` object:
```python
reg3 = regions.Regions([regions.CircleSkyRegion(center=SkyCoord('18:46:03.78', '-2:39:22.435', frame='icrs', unit=(u.h, u.deg)), radius=0.15*u.arcsec, visual=regions.RegionVisual({'color':'r'}))])
```
but this should be handled by the initializer.

larrybradley:
This PR allows region `meta` or `visual` values to be input as a `dict`, which is then converted by new descriptor classes to a `RegionMeta` or `RegionVisual` object.  The descriptor classes also add validation to ensure that `meta` and `visual` are either a `dict` or the appropriate object (e.g. `reg.visual = 'blue'` will now raise an error).

I consider this more as a new feature than a bug fix since the docs did not state that `dict` was previously an allowed input type.

Closes #461

### DIFF
--- a/regions/core/attributes.py
+++ b/regions/core/attributes.py
@@ -177,7 +177,7 @@ class RegionMetaDescr(RegionAttribute):
     """
     def __set__(self, instance, value):
         # RegionMeta subclasses dict
-        if type(value) == dict:  # pylint: disable=C0123
+        if isinstance(value, dict) and not isinstance(value, RegionMeta):
             value = RegionMeta(value)
         super().__set__(instance, value)
 
@@ -197,7 +197,7 @@ class RegionVisualDescr(RegionAttribute):
 
     def __set__(self, instance, value):
         # RegionVisual subclasses dict
-        if type(value) == dict:  # pylint: disable=C0123
+        if isinstance(value, dict) and not isinstance(value, RegionVisual):
             value = RegionVisual(value)
         super().__set__(instance, value)
 
--- a/regions/core/attributes.py
+++ b/regions/core/attributes.py
@@ -11,6 +11,7 @@ from astropy.units import Quantity
 import numpy as np
 
 from .pixcoord import PixCoord
+from .metadata import RegionMeta, RegionVisual
 
 __all__ = []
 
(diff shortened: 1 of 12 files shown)


regions/core/attributes.py
```diff
@@ -165,3 +166,42 @@ def _validate(self, value):
...
+class RegionMetaDescr(RegionAttribute):
+    """
+    Descriptor class for the region meta dictionary.
+
+    If input as a pure `dict`, it will be converted to a `RegionMeta`
+    object.
+    """
+    def __set__(self, instance, value):
+        # RegionMeta subclasses dict
+        if type(value) == dict:  # pylint: disable=C0123
```
keflavich, line 180: Is this intentionally not `isinstance(value, dict)`?  I think that can result in subclasses of dictionaries being allowed - which I think is something we want here.

regions/core/attributes.py
```diff
@@ -165,3 +166,42 @@ def _validate(self, value):
...
+    """
+    Descriptor class for the region visual dictionary.
+
+    If input as a pure `dict`, it will be converted to a `RegionVisual`
+    object.
+    """
+
+    def __set__(self, instance, value):
+        # RegionVisual subclasses dict
+        if type(value) == dict:  # pylint: disable=C0123
```
keflavich, line 200: ditto comment above

keflavich (changes requested):
lgtm, but I have a technical comment inline before I hit 'approve'.

regions/core/attributes.py
larrybradley: Yes, it's intentionally not `isinstance(value, dict)` because `RegionMeta` subclasses `dict`.  We only want to convert `dict` input into `RegionMeta`.  So it's either this or:
`if isinstance(value, dict) and not isinstance(value, RegionMeta):`.

larrybradley: But I wasn't thinking about *other* subclasses of dict.  I've updated this.  Thanks!

regions/core/attributes.py
larrybradley: Updated.

Issue #465, opened 2022-06-10
Inconsistent return type between SkyRegion.contains and PointSkyRegion.contains


registerrier:
The  implementation of `PointSkyRegion.contains()` always returns a single `bool` independently of the input `SkyCoord` shape.

The return type should be an array if the `SkyCoord.shape` is not scalar as is the case for the parent `SkyRegion.contains()`. 

This  behavior is correctly implemented for `PointPixelRegion` but not for `PointSkyRegion`.

See: https://github.com/astropy/regions/blob/9bfead1f6c62b20106dfc2cd545c8f85295ea95f/regions/shapes/point.py#L79

vs: https://github.com/astropy/regions/blob/9bfead1f6c62b20106dfc2cd545c8f85295ea95f/regions/shapes/point.py#L181

larrybradley:
@registerrier By design, `regions` only supports scalar shapes for all shape types, i.e. a single shape at a single position.  That premise was established by @astrofrog from the start.  Trying to initialize a `PointSkyRegion` raises an error: `ValueError: 'center' must be a scalar SkyCoord`.

PR #466, opened 2022-06-10
Remove PointSkyRegion.contains()

registerrier:
This PR solves issue #465 removing the overloaded `contains` method.
The test is corrected to compare the arrays themselves.


larrybradley:
As mentioned in #465, regions shape objects are always scalar by design.

Issue #469, opened 2022-07-14
`CompoundPixelRegion.plot()` needs center 


robelgeda:
Issue when trying to plot a compound region:

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Input In [66], in <cell line: 2>()
     68 chip_pixel_region = chip_region.to_pixel(wcs=wcs)
---> 69 chip_pixel_region.plot(fc='None', ec='blue', label=f"{region_catalog}", lw=4)

File ~/miniconda3/envs/astroconda/lib/python3.9/site-packages/regions/core/core.py:406, in PixelRegion.plot(self, origin, ax, **kwargs)
    403 if ax is None:
    404     ax = plt.gca()
--> 406 artist = self.as_artist(origin=origin, **kwargs)
    407 ax.add_artist(artist)
    409 return artist

File ~/miniconda3/envs/astroconda/lib/python3.9/site-packages/regions/core/compound.py:146, in CompoundPixelRegion.as_artist(self, origin, **kwargs)
    126 def as_artist(self, origin=(0, 0), **kwargs):
    127     """
    128     Return a matplotlib patch object for this region
    129     (`matplotlib.patches.PathPatch`).
   (...)
    144         A matplotlib patch object.
    145     """
--> 146     if (self.region1.center == self.region2.center
    147             and self.operator is op.xor):
    149         import matplotlib.patches as mpatches
    151         # set mpl_kwargs before as_artist is called on region1 and
    152         # region2

AttributeError: 'CompoundPixelRegion' object has no attribute 'center'
```

PR #470, opened 2022-07-15
WIP/proposal: plotting for compound artists

keflavich:
Fix for #469.  A little more documentation, and some testing, is needed.

@larrybradley what do you think of this general approach?  `.plot` stays transparent to users (though now can return lists), but `Composite` regions behave differently from other regions by default (except for annuli, which already have their own special case).

### DIFF
--- a/regions/core/compound.py
+++ b/regions/core/compound.py
@@ -123,6 +123,23 @@ class CompoundPixelRegion(PixelRegion):
 
         return mpath.Path(verts, codes)
 
+
+    def as_artists(self, origin=(0, 0), **kwargs):
+        """
+        Return a list of patches for the composite region.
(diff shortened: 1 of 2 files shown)


larrybradley (commented):
I think this approach may be the best we can do.  One issue is that the return patches will represent each component of the compound shape instead of single patch/path representing the outline of the compound shape.  But I think it would be challenging to implement that in a general way for all possible compound shapes.

keflavich:
Yeah, that's where I landed on this.  It's been a while since I worked on this PR - I think I was waiting for your go-ahead before doing any more work, but probably all that's needed are some tests & documentation?  (...which is twice as much as what's already here...)

PR #477, opened 2022-09-19
Test with Python 3.11 RC2

larrybradley:


.github/workflows/ci_tests.yml
```diff
@@ -33,6 +33,9 @@ jobs:
           - os: ubuntu-latest
             python: '3.10'
             tox_env: 'py310-test-alldeps'
+          - os: ubuntu-latest
+            python: '3.11.0-rc.2'
+            tox_env: 'py311-test-alldeps'
```
pllim, line 38: It won't pass with stable astropy (yet).

```suggestion
            tox_env: 'py311-test-devdeps'
```

larrybradley: thanks

pllim (approved):
Well, that was easy. Thanks! 😸

PR #486, opened 2022-10-27
Implement CircleSectorPixRegion

adonath:
This PR is a draft to implement a `CircleSectorPixRegion`. I needed this for a specific analysis, so I decided to implement it. Here is a notebook to illustrate that the bounding box and contains methods work correctly for all quadrants https://github.com/adonath/notebooks-public/blob/master/circle_sector_pix_region_illustration.ipynb

### DIFF
--- a/docs/shapes.rst
+++ b/docs/shapes.rst
@@ -59,6 +59,21 @@ Circle
     ...                                       outer_radius=5.2)
 
 
+`~regions.CircleSectorPixelRegion`
+
+.. code-block:: python
+
+    >>> from astropy.coordinates import SkyCoord
+    >>> from astropy import units as u
+    >>> from regions import PixCoord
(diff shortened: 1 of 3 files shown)


adonath:
@larrybradley is this a desired feature? Are there any guidelines with respect to adding new shapes to regions?

larrybradley:
@adonath This looks like a nice addition to me!  Implementing the "exact" mode for exact partial pixel overlaps could be tricky.

jzuhone:
I'd love to help with this. What do we need to get it across the finish line?

keflavich:
@adonath are you still working on this?  If the example works, maybe we can just merge (after resolving conflicts) and leave the unimplemented modes as open issues?

adonath:
@jzuhone @keflavich I have not made any progress since, but I would be happy rebase and open issues for the remaining tasks.

PR #503, opened 2023-04-28
Fix rectangle rotation corners test due to numpy sin/cos changes

larrybradley:
Due to an upstream change in numpy dev in the `sin` and `cos` functions (https://github.com/numpy/numpy/pull/23399), `np.sin(np.pi / 2)` no longer exactly equals 1.0.  👀 

This PR changes the test to use `allclose` instead of `equal`.

This should fix the currently-failing `devdeps` CI test.


pllim:
but_math_is_math

larrybradley:
All green again.

PR #511, opened 2023-09-07
Update nightly astropy-dev install

larrybradley:


larrybradley:
The remaining devdeps failures are due to NumPy 2.0 deprecations upstream in matplotlib.

PR #515, opened 2023-09-08
Remove unused array_compare tests

larrybradley:
In pytest >7.2 test functions can only return `None`.
Followup to https://github.com/astropy/regions/pull/480

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -105,7 +105,7 @@ norecursedirs = [
 astropy_header = true
 doctest_plus = 'enabled'
 text_file_format = 'rst'
-addopts = '--color=yes --doctest-rst'
+addopts = '--color=yes --doctest-rst --arraydiff'
 xfail_strict = true
 remote_data_strict = true
 filterwarnings = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -78,7 +78,7 @@ namespaces = false
 'regions' = [
     'CITATION.rst',
 ]
-'regions.shape.tests' = [
+'regions.shapes.tests' = [
     'reference/*.txt',
     'data/*.fits',
 ]
--- a/regions/shapes/tests/test_masks.py
+++ b/regions/shapes/tests/test_masks.py
@@ -42,7 +42,7 @@ def label(value):
                         for key, value in sorted(value.items()))
 
 
-@pytest.mark.array_compare(fmt='text', write_kwargs={'fmt': '%12.8e'})
+@pytest.mark.array_compare(file_format='text', write_kwargs={'fmt': '%12.8e'})
 @pytest.mark.parametrize(('region', 'mode'),
                          itertools.product(REGIONS, MODES), ids=label)
 def test_to_mask(region, mode):
--- a/regions/shapes/tests/test_masks.py
+++ b/regions/shapes/tests/test_masks.py
@@ -51,3 +51,4 @@ def test_to_mask(region, mode):
     except NotImplementedError:
         pytest.xfail()
     assert isinstance(mask, RegionMask)
+    return mask.data.astype(float)


astrofrog:
It's possible to fix this though, we fixed it in pytest-mpl and I need to copy over the fixes to pytest-arraydiff

larrybradley:
@astrofrog Great!  Let me know when there's a new `pytest-arraydiff` release with the fix.

astrofrog:
@larrybradley - could you try the latest dev version of pytest-arraydiff? Just merged a PR fixing this. Thanks!

larrybradley:
@astrofrog I've updated this PR to use `pytest-arraydiff` from `main`.  The good news is the tests do not fail.  But the bad news is that I changed some of the reference data  and even removed an entire file (see the last commit in this PR) and the tests still passed.  So it does not appear to be really compare the test output with the reference data.

pyproject.toml
```diff
@@ -39,7 +39,7 @@ all = [
     'shapely',
 ]
 test = [
-    'pytest-arraydiff',
+    'pytest-arraydiff @ git+https://github.com/astropy/pytest-arraydiff',
```
pllim, line 42: I released v0.6.0 so this is no longer necessary to pull from main
```suggestion
    'pytest-arraydiff',
```

regions/shapes/tests/test_masks.py
pllim, line 1: Did this used to work as intended in older pytest?

I look at https://github.com/astropy/pytest-arraydiff/blob/main/tests/test_pytest_arraydiff.py and none of the tests are stack with parametrize. And `reference_dir` is explicitly provided there.

pyproject.toml
larrybradley: done

regions/shapes/tests/test_masks.py
larrybradley: According to the README file (https://github.com/astropy/pytest-arraydiff#using), `reference` is the default directory name.

larrybradley:
I think I discovered the issue.  The `pytest` `--arraydiff` option was missing from the tox command.  For some reason if `--arraydiff` is not specified, `pytest` fails with errors complaining that the tests have a return value, e.g. "pytest.PytestReturnNotNoneWarning: Expected None, but regions/shapes/tests/test_masks.py::test_to_mask[poly-mode_subpixels-subpixels_5] returned array([[0.44, 0.4 , 0.2 , 0.04, 0.  , 0.  , 0.  ],"

larrybradley:
The `--arraydiff` option was missing from `pytest`.  Running `pytest` locally works, but strangely tox does not -- it cannot find the reference data.

Also, now if I do *not* add the `--arraydiff` option, `pytest` fails because the function does not return `None`.

pllim:
@larrybradley , does that mean pytest-arraydiff is working as expected or there is something else to be fixed? Hope you can clarify. Thanks!

larrybradley:
It seems `pytest-arraydiff` is working locally as expected when running `pytest` directly.  However, it fails when running the tests with `tox` (even though `tox` is running the same `pytest` command).  `tox` cannot find the reference data for comparison.  Is there some path setting or something that needs to be added with `tox`?

larrybradley:
This PR should be ready to go, but the (`tox`) tests are failing because the reference data is not found:  https://github.com/astropy/regions/actions/runs/6885867875/job/18730669107?pr=515

larrybradley:
I'm wondering now if the reference data isn't being packaged.

pllim:
Well, tox is supposedly isolated and whatever, so maybe you do need to explicitly point to the reference path even though it is supposedly the default location.

larrybradley:
@astrofrog -- Any idea why `tox` cannot find the `reference` data?  I took a look at `reproject` for reference and I don't see what's wrong here.

larrybradley:
I fixed the issue -- there was a typo in `pyproject.toml` `package-data`.  🤦

Issue #510, opened 2023-08-31
Allow Regions to be initialized without input


pllim:
```python
from regions import Regions
regs = Regions()  # Would be same as Regions([])
```

Motivation: This would make it possible to use Regions with `defaultdict`.

xref https://github.com/spacetelescope/jdaviz/pull/2410

Issue #509, opened 2023-08-31
Regions need to check inputs


pllim:
https://github.com/astropy/regions/blob/main/regions/core/regions.py

`Regions` does not have much input checking. For example, this actually works but it makes no sense:

```python
>>> from regions import Regions
>>> Regions([1,2,3])
<Regions([1, 2, 3])>
```

xref https://github.com/spacetelescope/jdaviz/pull/2410

Issue #508, opened 2023-08-31
Regions.extend should take another Regions


pllim:
https://github.com/astropy/regions/blob/501d12d5f5879c49413e20bed90a2d3eb725d5b7/regions/core/regions.py#L60-L70

Would be nice to be able to pass in another set of Regions without having to access its `.regions`. Example:

```python
regions_1 = Regions([r1, r2, r3])
regions_2 = Regions([r4, r5, r6])
regions_1.extend(regions_2)  # now regions_1 has r1-r6
```

It is kinda a shortcut for `regions_1.extend(regions_2.regions)`. Or do you think this is not worth it? Personally, I find `regions_1.extend(regions_2)` more intuitive.

xref https://github.com/spacetelescope/jdaviz/pull/2410

PR #527, opened 2023-09-20
Updates for Regions inputs

larrybradley:
This PR updates the `Regions` class as follows:

- allows initialization without any input, e.g., `Regions()`
- adds validation checking of input regions for the `Regions` object and its `append` and `extend` methods
- `Regions.extend` now accepts a `Regions` object

Closes: #508
Closes: #509
Closes: #510 

CC: @pllim


regions/core/regions.py
```diff
@@ -24,7 +24,13 @@ class Regions:
         The list of region objects.
     """
 
-    def __init__(self, regions):
+    def __init__(self, regions=(), /):
```
pllim, line 27: What is the `/` for?

pllim (approved):
LGTM. Thanks!

PR #548, opened 2024-04-09
Fix DS9 IO of Text regions

larrybradley:
This PR fixes an issue where writing Text regions (e.g., `TextPixelRegion`) to DS9 would drop the `text` field.  It also fixes an issue in the DS9 file reader/parser to have only one copy of the `text` field in the object instance (i.e., not duplicated in `meta`).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,10 @@ New Features
 Bug Fixes
 ---------
 
+- Fixed a bug where the text string of ``TextPixelRegion`` and
+  ``TextSkyRegion`` was dropped when serializing or writing
+  to a DS9 file. [#548]
+
 API Changes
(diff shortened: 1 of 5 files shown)


keflavich (approved):
I'm not sure I follow why `text` is a special parameter - is it not just normal metadata?  (I haven't looked closely at the unchanged source in a while, so maybe it's obvious)

larrybradley:
For text regions, DS9 stores the text value the same was as additional metadata.  It's not included in the shape definition (only the coords are).  So for Text regions, we need to explicitly add "text" to the metadata when writing, and then remove it when reading (it gets stored in the class attribute) -- that prevents two copies of the same thing and them going out of sync.

PR #587, opened 2025-02-11
MNT: Replace ubuntu-20.04 with ubuntu-22.04

pllim:
As recommended by actions/runner-images#11101 , this PR replaces the deprecated ubuntu-20.04 runner with a newer version. This also replaces RTD workflow, if applicable, to future-proof the build in case it follows suit.

This is an automated update made by the `batchpr` tool :robot: - feel free to close if it doesn't look good! You can report issues to @pllim.

[:ghost:](https://github.com/pllim/playpen/issues/63)

### DIFF
--- a/.github/workflows/ci_cron_weekly.yml
+++ b/.github/workflows/ci_cron_weekly.yml
@@ -67,7 +67,7 @@ jobs:
     # we include them just in the weekly cron. These also serve as a test
     # of using system libraries and using pytest directly.
 
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     name: More architectures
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_cron_weekly.yml
```diff
@@ -67,7 +67,7 @@ jobs:
     # we include them just in the weekly cron. These also serve as a test
     # of using system libraries and using pytest directly.
 
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     name: More architectures
```
pllim, line 71: Hmm the aarch64 job doesn't need emulator anymore but it is out of scope here. See https://github.com/astropy/photutils/pull/1997

pllim:
@larrybradley , feel free to cancel the aarch64 job and just merge since you are going to refactor that one anyway. Thanks!

PR #623, opened 2025-09-03
fix unit conversion error in wcs_helpers

### LINKED ISSUE #622: `pixel_scale_angle_at_skycoord` broken with updated numpy/astropy versions
The latest numpy/astropy versions propagate units through trigonometric functions, e.g. here
https://github.com/astropy/regions/blob/72567c480508dadfd7553a167b6f19fd558741e5/regions/_utils/wcs_helpers.py#L66

By using the `*` operator, the arctangent is converted into radian^2 which breaks the unit conversion.

One fix is to swap from the `*` operator to `<<` (which would be advisable anyway).

Here's a snippet showing the behavior
```py
import numpy as np
import astropy.units as u

np.arctan2(1 << u.pixel, 2 << u.pixel)
# output: 0.46364761 rad
```

settwi:
Fixes #622

### DIFF
--- a/regions/_utils/wcs_helpers.py
+++ b/regions/_utils/wcs_helpers.py
@@ -62,7 +62,7 @@ def pixel_scale_angle_at_skycoord(skycoord, wcs, offset=1 * u.arcsec):
 
     dx = x_offset - x
     dy = y_offset - y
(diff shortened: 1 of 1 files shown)


larrybradley:
Closing, see https://github.com/astropy/regions/issues/622#issuecomment-3300892424.

PR #628, opened 2025-10-30
Don't overwrite input meta/visual for compound sky regions

rosteen:
I noticed that while `CompoundPixelRegion` uses the input value for `meta` and `visual` if they're specified, `CompoundSkyRegion` overwrites them with `RegionMeta()` and `RegionVisual()`. This looks like a bug, so this PR changes the code to actually use the input values on init.


keflavich (approved):
this looks right to me

PR #661, opened 2026-05-02
Upgrade PyPI CI publishing to use Trusted Publishing

lmmx:
- Background: https://github.com/astropy/specutils/issues/1323

Migrates PyPI publishing from a long-lived API token to [Trusted Publishing](https://docs.pypi.org/trusted-publishers/) (OIDC), motivated by recent supply chain attacks ([litellm](https://blog.pypi.org/posts/2026-04-02-incident-report-litellm-telnyx-supply-chain-attack/), [lightning](https://cybersecuritynews.com/python-package-lightning-hacked/)).

The publishing step uses [a reusable workflow](https://github.com/OpenAstronomy/github-actions-workflows/blob/v2.6.3/.github/workflows/publish.yml) from the OpenAstronomy org, and since you can't pass OIDC tokens across orgs, the workaround (documented [here](https://github-actions-workflows.openastronomy.org/en/latest/trusted_publishing.html)) is to instead set `upload_to_pypi: false` and `save_artifacts: true` then publish the stored artifact with the `pypa/gh-action-pypi` action. I followed [this example](https://github.com/astropy/astropy/blob/80447a4334ba617c04d5918b66d80793f202c52e/.github/workflows/publish.yml#L77-L94) from the astropy package repo.

The code changes here require some further (trivial) setup on the PyPI-side. Specifically, the PyPI admin (not just maintainer) needs to register the TP on PyPI at https://pypi.org/manage/project/regions/settings/publishing/

- Owner: `astropy`
- Repo: `regions`
- Workflow: `publish.yml`
- Environment: `pypi`

The `pypi_token` secret should be deleted from the repo secrets and can be invalidated on PyPI too.

(Not sure who is the PyPI admin, the maintainers are listed as @keflavich @larrybradley @astrofrog)

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -29,10 +29,6 @@ permissions:
 
 jobs:
   build_and_publish:
-    # This job builds the wheels and publishes them to PyPI for all
-    # tags, except those ending in ".dev". For PRs with the "Build all
-    # wheels" label, wheels are built, but are not uploaded to PyPI.
-
     permissions:
       contents: none
 
@@ -40,8 +36,8 @@ jobs:
 
     if: (github.repository == 'astropy/regions' && (github.event_name == 'push' ||  github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' || contains(github.event.pull_request.labels.*.name, 'Build all wheels')))
     with:
-      # We upload to PyPI for all tag pushes, except tags ending in .dev
-      upload_to_pypi: ${{ startsWith(github.ref, 'refs/tags/') && !endsWith(github.ref, '.dev') && (github.event_name == 'push' || github.event_name == 'workflow_dispatch') }}
+      upload_to_pypi: false
+      save_artifacts: true
 
       test_extras: test
       test_command: pytest -p no:warnings --pyargs regions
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -63,5 +59,28 @@ jobs:
...
+    # This job publishes the built wheels to PyPI for all tags, except
+    # those ending in ".dev". For PRs with the "Build all wheels" label,
+    # wheels are built, but are not uploaded to PyPI.
+    permissions:
+      id-token: write
+    environment:
+      name: pypi
+      url: https://pypi.org/project/regions
+    # We upload to PyPI for all tag pushes, except tags ending in .dev
+    if: startsWith(github.ref, 'refs/tags/') && !endsWith(github.ref, '.dev') && (github.event_name == 'push' || github.event_name == 'workflow_dispatch')
```
larrybradley, line 74: Is it worth adding `github.repository == 'astropy/regions'` here too?  I know a user fork won't actually be published, but it could cause wasted CI resources/runs if a user pushes a tag to a fork.

larrybradley (approved):
Thanks, @lmmx.  I've removed the `pypi_token` and set up Trusted Publishing on PyPI.

@astrofrog -- do you also want to review?

.github/workflows/publish.yml
lmmx: No, because the `build_and_publish` job is already gated by that (on line 37), and the upload job needs that job to have succeeded first, so is effectively already gated

larrybradley: @lmmx My understanding is that a *skipped* job (because of the `build_and_publish` conditional) will report its status as "Success".  From the [GHA docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks):

 > A job that is skipped will report its status as "Success".
 
 Is something more explicit like `needs.build_and_publish.result == 'success'` needed in the `upload` job?

larrybradley: I'll update this in a follow up.

PR #674, opened 2026-06-02
Strip whitespace from FITS region SHAPE values (fixes #637)

### LINKED ISSUE #637: Strip white space when reading regions from FITS files
When trying to read region files downloaded from the Chandra Source Catalog using CSCView, I ran into multiple issues. 

One is that the extension names in the fits file are "SRCREG" and "BKGREG", but `Regions.read` only accepts "REGION". It may be a good idea to add an argument to allow the user to chose which extension to use. 

The second issue is that the region shape had trailing white space when read by `parse_table`, preventing it from being recognized. Perhaps this is an issue with encoding on Windows. I added `.strip()` to line 79 in `read.py` as a quick fix.

gaoflow:
## Description

Reading a FITS region file whose `SHAPE` column values carry leading or trailing whitespace raises a spurious `FITSParserError`. This was reported in #637 for region files exported from the Chandra Source Catalog via CSCView (where, likely due to fixed-width character-column padding and/or cross-platform encoding, the shape token comes back with trailing whitespace).

`get_shape()` matched the raw column value against the list of valid shape names:

```python
shape = region_row['SHAPE'].lower()
if shape[0] == '!':
    ...
```

so a value such as `'circle  '` failed the `valid_shapes` membership check and raised:

```
FITSParserError: 'circle  ' is not a valid FITS region shape
```

## Fix

Strip surrounding whitespace from the `SHAPE` value before matching, and use `str.startswith('!')` instead of indexing `shape[0]` for the exclude prefix so that an empty value after stripping no longer raises an `IndexError`.

```python
shape = region_row['SHAPE'].strip().lower()
if shape.startswith('!'):
    ...
```

The change is a no-op for already-clean values, so existing files are unaffected.

## Reproduce

```python
import astropy.units as u
from astropy.table import QTable
from regions import Regions

tbl = QTable([[1] * u.pix, [2] * u.pix, [3] * u.pix, ['circle  ']],
             names=('X', 'Y', 'R', 'SHAPE'))
Regions.parse(tbl, format='fits')   # FITSParserError before, parses after
```

## Tests

Adds `test_shape_with_whitespace` covering a padded shape name and a whitespace-wrapped `'!'`-excluded shape. The new test fails on `main` (raising `FITSParserError`) and passes with the fix; the full `regions/io/fits` test suite passes (7 passed).

Fixes #637.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -54,6 +54,11 @@ New Features
 Bug Fixes
 ---------
 
+- Fixed reading FITS region files whose ``SHAPE`` column values contain
+  leading or trailing whitespace (e.g. from fixed-width character-column
+  padding or cross-platform encoding quirks), which previously raised a
+  spurious ``FITSParserError``. [#674]
+
 API Changes
 -----------
 
--- a/regions/io/fits/read.py
+++ b/regions/io/fits/read.py
@@ -76,8 +76,15 @@ def get_shape(region_row):
         shape = 'point'
         return shape, include
 
-    shape = region_row[shape_key].lower()
-    if shape[0] == '!':
+    # Strip surrounding whitespace before matching the shape name. FITS
+    # character columns are fixed-width and space-padded, and some writers
+    # (or cross-platform/encoding quirks) leave trailing whitespace on the
+    # SHAPE value, which would otherwise fail the ``valid_shapes`` check
+    # below and raise a spurious FITSParserError (see GH #637). Using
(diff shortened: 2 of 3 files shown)


keflavich (approved):
Looks like a real problem & good solution.  The comments are a bit on the verbose side, but descriptive and harmless.

larrybradley (approved):
Thanks, @gaoflow

PR #676, opened 2026-06-05
Fix pixel <-> sky conversions for edge cases

larrybradley:
This PR includes 3 fixes to the region `to_pixel` and `to_sky` conversions for a few edge cases:

* Fixed the ``to_sky`` and ``to_pixel`` conversions for elliptical and rectangular regions (and their annulus variants) so that the orientation is correct for a WCS with a flipped parity (e.g., North down, East left). Previously, such regions were mirrored about the x-axis.

* Fixed the local WCS Jacobian computation used by the ``to_sky`` and ``to_pixel`` methods so that it is well-conditioned at the celestial poles and across the RA = 0/360 degree wraparound.

* Fixed rectangular regions to use the SVD-based ellipse ``to_sky`` and ``to_pixel`` conversions so that they round-trip correctly with a sheared WCS. [#676]

Based on https://github.com/astropy/photutils/pull/2286 and https://github.com/astropy/photutils/pull/2288


settwi:
Nice, this is a helpful fix to have for oriented boxes. For [SDO/AIA](https://sdo.gsfc.nasa.gov/) images the axes are (sometimes?) flipped, so for boxing outflows this is an important bugfix. Thanks.

PR #677, opened 2026-06-05
region __eq__ method uses all_close, like PixCoord

### LINKED ISSUE #678: Loosen restriction on SkyCoord centers when comparing regions
When checking equivalency, PixCoord tests that coordinates are close-enough, while SkyCoord tests that coordinates match exactly. This becomes and issue when checking for region equivalency, the restrictions are looser for PixelRegions which can have centers that have small precision differences, where SkyRegions much match exactly (except for angle, where Region.\_\_eq__ uses all close).

cshanahan1:
This PR updates Region's \_\_eq__ method to use allclose instead of checking for exact equality. 

This is presently how equivalency is checked for PixCoord, so there was a mismatch between looser requirements for closeness of pixel centers vs angles - angles were required to match exactly, while centers could have small numerical differences and still be equal.

This requires a follow up workaround for SkyCoord (see https://github.com/astropy/regions/issues/678)

Fixes #678

### DIFF
--- a/regions/core/core.py
+++ b/regions/core/core.py
@@ -166,20 +166,53 @@ class Region(abc.ABC):
     @staticmethod
     def _is_close(val1, val2):
         """
-        Helper to compare PixCoord, Quantity, SkyCoord, and other parameters.
+        Helper to compare PixCoord, Quantity, SkyCoord, and other
+        parameters.
+
+        For SkyCoord objects, both coordinates are converted to ICRS and
+        their RA/Dec values are compared using `numpy.allclose` with
+        default tolerances, enabling tolerance-based equivalency similar
+        to PixCoord and Quantity parameters.
+
+        SkyCoord objects must have equivalent frames to be considered
+        close.
+
+        Parameters
+        ----------
+        val1, val2 : object
+            Values to compare.
+
+        Returns
+        -------
+        bool
+            True if values are considered close/equal, False otherwise.
         """
         # PixCoord has its own __eq__ with built-in tolerances
         if isinstance(val1, PixCoord):
             return val1 == val2
 
+        # SkyCoord comparison using RA/DEC allclose
+        if isinstance(val1, SkyCoord) and isinstance(val2, SkyCoord):
+            try:
+                # Frames must be equivalent, convert to ICRS for comparison
+                if val1.frame.is_equivalent_frame(val2.frame):
+                    ra_close = np.allclose(val1.icrs.ra.deg,
+                                           val2.icrs.ra.deg)
+                    dec_close = np.allclose(val1.icrs.dec.deg,
+                                            val2.icrs.dec.deg)
+                    return ra_close and dec_close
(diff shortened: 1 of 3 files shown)


larrybradley:
pre-commit.ci autofix

regions/core/core.py
```diff
@@ -185,17 +185,28 @@ def __eq__(self, other):
...
+            self_val = getattr(self, param)
+            other_val = getattr(other, param)
+
+            # compare PixCoord directly, the PixCoord __eq__ method also uses
+            # np.allclose with the default tolerances
+            if isinstance(self_val, PixCoord) and self_val != other_val:
+                return False
+            else:
+                try:
+                    if not np.allclose(self_val, other_val):  # also handles comparible units
```
cshanahan1: @ the reviewer, turns out np.allclose also works with quantities even if they're in different units as long as they're equivalent, which I didn't know. I am using np.allclose instead of u.allclose to be consistent with the default tolerance PixCoord uses.

regions/core/core.py
```diff
@@ -185,17 +185,28 @@ def __eq__(self, other):
...
+                except TypeError:
+                    # fallback direct comparison for empty dicts, or for
+                    # SkyCoord comparisons which return false when frames
+                    # are not equivalent
+                    try:
+                        if np.any(self_val != other_val):
+                            return False
+                    except TypeError:
+                        return False
         return True
```
larrybradley: How about we use a small `_is_close` helper function to simplify the `__eq__` logic?  (this also avoids the linter complaining about too many `return` statements in one method)

Perhaps something like this?

```python
@staticmethod
def _is_close(val1, val2):
    """
    Helper to compare PixCoord, Quantity, SkyCoord, and other parameters.
    """
    # PixCoord has its own __eq__ with built-in tolerances
    if isinstance(val1, PixCoord):
        return val1 == val2

    try:
        # np.allclose handles Quantities with convertible units
        return np.allclose(val1, val2)
    except (TypeError, ValueError):
        try:
            # Fallback for SkyCoord (which may return False if frames differ)
            # or dicts/objects where np.allclose fails.
            return not np.any(val1 != val2)
        except (TypeError, ValueError):
            return False

def __eq__(self, other):
    """
    Equality operator for Region.

    All Region properties are compared for strict equality except
    for Quantity parameters, which allow for different units if they
    are directly convertible.
    """
    if not isinstance(other, self.__class__):
        return False

    # Define the parameters to check
    params = list(self._params) + ['meta', 'visual']
    other_params = list(other._params) + ['meta', 'visual']

    # Check that both have identical parameter sets
    if params != other_params:
        return False

    # Check that all parameter values are close
    return all(self._is_close(getattr(self, p), getattr(other, p)) for p in params)
```

larrybradley:
pre-commit.ci autofix

larrybradley (approved):
Thanks, @cshanahan1!  I pushed a commit to address the SkyCoord allclose comparison.

PR #689, opened 2026-06-23
Rename include_boundary_distortions to boundary_distortions

larrybradley:
This PR renames the `include_boundary_distortions` keyword to `boundary_distortions`.  The shorter name still preserves this boolean keyword meaning.


larrybradley:
I plan to make a new release after this PR.

PR #700, opened 2026-08-23
Fix doubled comma in CRTF output when range/corr is the only metadata

### LINKED ISSUE #322: Meta data output in CRTF file
When using the global coordinates in CRTF files, in the output line two commas appear between the shape definition and the meta strings, like so:

`ellipse[[279.486483deg, -20.683327deg], [0.004609deg, 0.002977deg], 100.410911deg] , ,corr=XX`

Since all the meta strings are defined with a preceding comma (which makes sense). Still, this is a problem when trying to specify meta data.

preller:
This PR fixes the CRTF writer emitting a doubled comma (`, ,`) when `range` or `corr` is the only metadata of a region, which makes the output unparseable by CASA. Relevant when CRTF files are generated programmatically for machine-readable pipelines. The metadata items are now collected in a list and joined once, so an empty separator cannot appear; this also removes the existing special case that guarded the `coord=` key against the same problem. A regression test is included. Fixes #322

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,9 @@ New Features
 Bug Fixes
 ---------
 
+- Fixed a doubled comma in the serialized CRTF output when ``range`` or
+  ``corr`` is the only metadata of a region. [#700]
+
 API Changes
 -----------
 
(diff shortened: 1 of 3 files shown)


keflavich (approved):
Good catch, assuming this passes CI, lgtm.

larrybradley (approved):
Thanks, @preller

